PR #16, opened 2025-06-17
update EasyBuild hooks to only use 12 cores when building QuantumESPRESSO for A64FX

boegel:


ocaisa:
#17 is being deployed so closing this

PR #21, opened 2025-06-24
filter whiteout files with `find` (instead of using `grep`) in `create_tarball.sh`

bedroge:
This should solve the issues in the tarball creation steps that we see in for instance https://github.com/EESSI/software-layer/pull/1131. There are some emtpy directories in the `init` folder, and then the `grep` returns 1. By adding `|| true` (we already did that for a few other `grep` commands) it will always return 0.

edit: instead of adding `|| true` to the `grep` command, we decided to add an option to the `find` command for excluding whiteout files.

### DIFF
--- a/create_tarball.sh
+++ b/create_tarball.sh
@@ -85,8 +85,7 @@ for subdir in ${cpu_arch_subdir} ${cpu_arch_subdir}/accel/${accel_subdir}; do
         # installation directories), the procedure will likely not work.
         for package_version in $(cat ${module_files_list}); do
             echo "handling ${package_version}"
-            ls -d ${eessi_version}/software/${os}/${subdir}/software/${package_version} \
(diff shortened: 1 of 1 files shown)


trz42 (approved):
Looks good.

An alternative way might be to use plain `find` with an additional condition, e.g.,
```bash
find /path/to/upper -type f ! -name '.wh.*'
```

Anyhow, up to you to decide if that's worth another change or we go with your change.

bedroge:
> Looks good.
> 
> An alternative way might be to use plain `find` with an additional condition, e.g.,
> 
> ```shell
> find /path/to/upper -type f ! -name '.wh.*'
> ```
> 
> Anyhow, up to you to decide if that's worth another change or we go with your change.

No strong preference, but I changed the `find` commands.

trz42 (approved):
Nice!

PR #9, opened 2025-06-14
add support to `eessi_container.sh` for using `unionfs` overlay tool

boegel:
Default behavior is left unchanged.

`unionfs` looks promising, haven't observed the [`Permission denied` issues](https://github.com/EESSI/software-layer/issues/556) we've been seeing when creating overlay with `fuse-overlays`, but it needs to be battle-tested before we jump ship.

Tested with:
```
./eessi_container.sh --container docker://ghcr.io/eessi/build-node:debian12 --overlay-tool unionfs --access rw
```
(see also https://github.com/EESSI/filesystem-layer/pull/226)

PR #22, opened 2025-06-24
Create for utils function nvidia-smi check

laraPPr:


boegel:
@laraPPr Ignore the failing CI check for now, I'm trying to fix that in #9 ...

ocaisa:
I think you can have a couple of tests for this, there's mocking for `nvidia-smi` in https://github.com/laraPPr/software-layer-scripts/blob/utils_function_nvidia/.github/workflows/tests_link_nvidia_host_libraries.yml#L37 .

ocaisa (changes requested):

bot/test.sh
```diff
@@ -228,17 +228,8 @@ TEST_STEP_ARGS+=("--extra-bind-paths" "/sys/fs/cgroup:/hostsys/fs/cgroup:ro")
 if command_exists "nvidia-smi"; then
     # Accept that this may fail
     set +e
-    nvidia-smi --version
-    ec=$?
+    check_nvidia-smi_installation
```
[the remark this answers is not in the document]
laraPPr: here the problem is that `TEST_STEP_ARGS+=("--nvidia" "run")` will not be set but instead `${EESSI_PREFIX}/scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh` will run. Gonna test this pr thoroughly in https://github.com/EESSI/software-layer-scripts/pull/49

bot/build.sh
```diff
@@ -220,25 +220,18 @@ BUILD_STEP_ARGS+=("--save" "${TARBALL_TMP_BUILD_STEP_DIR}")
...
+# Accept that this may fail
+set +e
+verify_nvidia-smi
+ec=$?
+if [ ${ec} -eq 0 ]; then
+    BUILD_STEP_ARGS+=("--nvidia" "all")
+elif [ ${ec} -eq 1 ]; then
+    BUILD_STEP_ARGS+=("--nvidia" "install")
+elif [  ${ec} -eq 2  ]; then 
     BUILD_STEP_ARGS+=("--nvidia" "install")
```
laraPPr, line 232: This is now exactly mimicking the behavior but I'm not sure this is correct. 
This is now set in this case `No 'nvidia-smi' found, no available GPU but allowing overriding this check`. But now it is always set or am I missing some logic?

bot/build.sh
```diff
@@ -220,25 +220,18 @@ BUILD_STEP_ARGS+=("--save" "${TARBALL_TMP_BUILD_STEP_DIR}")
...
-    echo "No 'nvidia-smi' found, no available GPU but allowing overriding this check"
+# Accept that this may fail
+set +e
+verify_nvidia-smi
+ec=$?
+if [ ${ec} -eq 0 ]; then
+    BUILD_STEP_ARGS+=("--nvidia" "all")
+elif [ ${ec} -eq 1 ]; then
+    BUILD_STEP_ARGS+=("--nvidia" "install")
+elif [  ${ec} -eq 2  ]; then
```
ocaisa, line 231:
```suggestion
if verify_nvidia-smi; then
    BUILD_STEP_ARGS+=("--nvidia" "all")
else
    # We don't have a GPU but we still want to do a GPU installation
```
Why bother handling the different modes if what you want to do is the same?

bot/build.sh
```diff
@@ -220,25 +220,18 @@ BUILD_STEP_ARGS+=("--save" "${TARBALL_TMP_BUILD_STEP_DIR}")
...
+verify_nvidia-smi
+ec=$?
+if [ ${ec} -eq 0 ]; then
+    BUILD_STEP_ARGS+=("--nvidia" "all")
+elif [ ${ec} -eq 1 ]; then
+    BUILD_STEP_ARGS+=("--nvidia" "install")
+elif [  ${ec} -eq 2  ]; then 
     BUILD_STEP_ARGS+=("--nvidia" "install")
 fi
+set -e
```
ocaisa, line 234: Shouldn't be necessary to capture these if you are handling return codes for a function

bot/test.sh
```diff
@@ -225,21 +225,14 @@ fi
...
-        echo "If, and only if, the current node actually does contain Nvidia GPUs, this should be considered an error."
-    fi
+# Accept that this may fail
+set +e
+verify_nvidia-smi
+ec=$?
+if [ ${ec} -eq 0 ]; then
+    TEST_STEP_ARGS+=("--nvidia" "run")
 fi
+set -e
```
ocaisa, line 235:
```suggestion
if verify_nvidia-smi; then
    TEST_STEP_ARGS+=("--nvidia" "run")
fi
```

scripts/utils.sh
```diff
@@ -147,3 +147,22 @@ function get_ipv4_address {
     echo "${hipv4}"
     return 0
 }
+
+function verify_nvidia-smi {
+    if command_exists "nvidia-smi"; then
+        nvidia-smi --version
+        ec=$?
+        if [ ${ec} -eq 0 ]; then
+            echo "Command 'nvidia-smi' found. Installing NVIDIA drivers for use in prefix shell..."
```
ocaisa, line 156: echo is no longer accurate
```suggestion
            echo "Command 'nvidia-smi' found..."
```

scripts/utils.sh
```diff
@@ -147,3 +147,22 @@ function get_ipv4_address {
...
+	    return 0
+        else
+            echo "Warning: command 'nvidia-smi' found, but 'nvidia-smi --version' did not run succesfully."
+            echo "This script now assumes this is NOT a GPU node."
+            echo "If, and only if, the current node actually does contain Nvidia GPUs, this should be considered an error."
```
ocaisa, line 161:
```suggestion
```

scripts/utils.sh
```diff
@@ -147,3 +147,22 @@ function get_ipv4_address {
...
+	    return 1
+	fi
+    else
+        echo echo "No 'nvidia-smi' found, no available GPU but allowing overriding this check"
```
ocaisa, line 165:
```suggestion
        echo "No 'nvidia-smi' found, no available GPU" 
```

EESSI-install-software.sh
```diff
@@ -315,19 +310,17 @@ else
 fi
 
 # Install NVIDIA drivers in host_injections (if they exist)
-if command_exists "nvidia-smi"; then
+# Accept that this may fail
+set +e
+verify_nvidia-smi
+ec=$?
+if [ ${ec} -eq 0 ]; then
+    export LD_LIBRARY_PATH="/.singularity.d/libs:${LD_LIBRARY_PATH}"
```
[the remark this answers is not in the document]
ocaisa: I think you could have this in the function itself when you call nvidia-smi, if it is needed it is required to make nvidia-smi work (and if it is not needed it does no harm anyway...unless it is exported!)

scripts/utils.sh
```diff
@@ -147,3 +147,22 @@ function get_ipv4_address {
     echo "${hipv4}"
     return 0
 }
+
+function verify_nvidia-smi {
+    if command_exists "nvidia-smi"; then
+        nvidia-smi --version
```
ocaisa, line 153:
```suggestion
        # We are careful here in case we are running in a container and LD_LIBARY_PATH has been wiped
        LD_LIBRARY_PATH="/.singularity.d/libs:${LD_LIBRARY_PATH}" nvidia-smi --version
```

EESSI-install-software.sh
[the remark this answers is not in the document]
laraPPr: I'm not sure why it is their. I only found it here https://github.com/EESSI/software-layer-scripts/blob/ce4f213e1dc093d3c4eab4ebca5c04cd56be23e6/EESSI-install-software.sh#L319 so that is why I added it this way but if it can't cause any harm in the build.sh and test.sh scrits I can add it to the function.

scripts/utils.sh
```diff
@@ -147,3 +147,22 @@ function get_ipv4_address {
     echo "${hipv4}"
     return 0
 }
+
+function verify_nvidia-smi {
```
ocaisa, line 151: I think the function name is misleading. What is this really doing? It is checking if there is an nvidia gpu is available. nvidia-smi is the tool we use to do that, but it is not the important thing. I think a better name is `nvidia_gpu_available`, and either it is or it isn't (no need for anything other than 0 or 1)

bot/build.sh
ocaisa: In our scripts this is fine, it is a once only thing, once things are installed they are not reinstalled.

ocaisa (changes requested):
This looks good, a CI check would be great. You can do it in https://github.com/EESSI/software-layer-scripts/blob/main/.github/workflows/tests_link_nvidia_host_libraries.yml, check once before `nvidia-smi` is mocked that it returns non-zero, and once after it has been mocked

ocaisa:
This looks good, can you sync it with `main`, I think the broken CI should have already been fixed there since Friday

PR #23, opened 2025-06-24
add workflow scripts for filtering and checking easystacks

laraPPr:


ocaisa:
@laraPPr This isn't the right place for these I would say, you want them in the `software-layer`

laraPPr:
I'm gonna close this. I'll add the scripts for now in the software-layer. And when I have time enough flesh the scripts out enough to create an action.

PR #26, opened 2025-07-01
use `unionfs` overlay tool (+ debian12 build container that includes it) by default

boegel:
Follow-up for:
* #9

As discussed during today's EESSI support sync meeting, let's change the default to use `unionfs` rather than `fuse-overlayfs` to create a writable overlay for building software to be ingested into EESSI.

This should circumvent the problems we have with `fuse-overlayfs` when doing rebuilds, see also https://github.com/EESSI/software-layer/issues/556 and https://github.com/EESSI/software-layer/issues/1103

### DIFF
--- a/.github/workflows/test_eessi_container_script.yml
+++ b/.github/workflows/test_eessi_container_script.yml
@@ -41,7 +41,7 @@ jobs:
         - name: Test eessi_container.sh script
           run: |
               test_cmd="cat /etc/os-release"
-              out_pattern="Debian GNU/Linux 11"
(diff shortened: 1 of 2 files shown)


bedroge:
Looks like some values related to the Debian version are hardcoded in the CI:
https://github.com/EESSI/software-layer-scripts/blob/main/.github/workflows/test_eessi_container_script.yml#L44
https://github.com/EESSI/software-layer-scripts/blob/main/.github/workflows/test_eessi_container_script.yml#L119

PR #31, opened 2025-07-08
Remove removal step from `bot/build.sh`

bedroge:
Also removes the `EESSI-remove.sh` script. This basically reverts the changes from https://github.com/EESSI/software-layer/pull/488, which were implemented as a workaround to rebuild software with fuse-overlayfs. With unionfs we no longer have these issues, so we can get rid of this step. By doing this, we no longer depend on fakeroot, which was causing issue on our grace and a64fx nodes.

In order to test this, I'll add a dummy easystack and trigger some test builds, and then remove the dummy easystacks again.


bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:x86_64/amd/zen2

bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:x86_64/amd/zen2

bedroge:
Those builds are probably failing because the attr sources are currently unavailable, had the same issue on the AWS cluster....

bedroge:
Okay, that worked fine. This PR should be ready, but the dummy easystacks need to be removed before merging this. I'll do that after this has been approved, so we can still test additional changes if required.

ocaisa (changes requested):
PR looks good, will approve once the easystack and `EESSI-extend` changes are removed

ocaisa (approved):

PR #36, opened 2025-07-10
Fix test accelerator override

casparvl:
I noticed [here](https://github.com/casparvl/software-layer/pull/1#issuecomment-3053068888) that in the test step, a different prefix (.../cc90) was being used than in the build step (.../cc80). That's clearly not what we want: we want to test what we build.


bedroge (approved):
Lgtm, same logic is used in `bot/build.sh`.

PR #41, opened 2025-07-24
Allow end users to configure EESSI module to be sticky

ocaisa:


bedroge (approved):
Smart, this is really nice to have!

bedroge:
Tarballs ingested.

PR #45, opened 2025-07-30
Add settings for EasyBuild 5+ to EESSI-extend

ocaisa:
@boegel Would be great if you can review this and see if I missed anything (and if my choices are "correct")

### DIFF
--- a/EESSI-extend-easybuild.eb
+++ b/EESSI-extend-easybuild.eb
@@ -98,21 +98,26 @@ if (eessi_accelerator_target ~= nil) then
   end
 end
 
+-- Some environment variables affect behaviour, let's gather them once
+local eessi_cvmfs_install = os.getenv("EESSI_PROJECT_INSTALL") ~= nil
+local eessi_site_install = os.getenv("EESSI_SITE_INSTALL") ~= nil
+local eessi_project_install = os.getenv("EESSI_PROJECT_INSTALL") ~= nil
+local eessi_user_install = os.getenv("EESSI_USER_INSTALL") ~= nil
+
 -- Use an installation prefix that we _should_ have write access to
-if (os.getenv("EESSI_CVMFS_INSTALL") ~= nil) then
+if (essi_cvmfs_install) then
   -- Make sure no other EESSI install environment variables are set
-  if ((os.getenv("EESSI_SITE_INSTALL") ~= nil) or (os.getenv("EESSI_PROJECT_INSTALL") ~= nil) or (os.getenv("EESSI_USER_INSTALL") ~= nil)) then
+  if (eessi_site_install or eessi_project_install or eessi_user_install) then
     LmodError("You cannot use EESSI_CVMFS_INSTALL in combination with any other EESSI_*_INSTALL environment variables")
   end
-  eessi_cvmfs_install = true
   easybuild_installpath = os.getenv("EESSI_SOFTWARE_PATH")
   -- enforce accelerator subdirectory usage for CVMFS installs (only if an accelerator install is requested)
   if (eessi_accelerator_target ~= nil) and (cuda_compute_capability ~= nil) and (os.getenv("EESSI_ACCELERATOR_INSTALL") ~= nil) then
       easybuild_installpath = pathJoin(easybuild_installpath, eessi_accelerator_target)
   end
-elseif (os.getenv("EESSI_SITE_INSTALL") ~= nil) then
+elseif eessi_site_install then
   -- Make sure no other EESSI install environment variables are set
-  if ((os.getenv("EESSI_PROJECT_INSTALL") ~= nil) or (os.getenv("EESSI_USER_INSTALL") ~= nil)) then
+  if (eessi_project_install or eessi_user_install) then
     LmodError("You cannot use EESSI_SITE_INSTALL in combination with any other EESSI_*_INSTALL environment variables")
   end
   easybuild_installpath = os.getenv("EESSI_SITE_SOFTWARE_PATH")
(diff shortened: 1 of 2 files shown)


EESSI-extend-easybuild.eb
```diff
@@ -196,10 +201,30 @@ elseif (project_modulepath ~= nil) then
...
+-- Set environment variables that are EasyBuild version specific
+if convertToCanonical(easybuild_version) > convertToCanonical("4") then
+  setenv ("EASYBUILD_STRICT_RPATH_SANITY_CHECK", "1")
+  setenv ("EASYBUILD_CUDA_SANITY_CHECK_ERROR_ON_FAILED_CHECKS", "1")
+  setenv ("EASYBUILD_ENV_FOR_SHEBANG", "1")
+  setenv ("EASYBUILD_FAIL_ON_MOD_FILES_GCCCORE", "1")
+  setenv ("EASYBUILD_LOCAL_VAR_NAMING_CHECK", "error")
+  -- Set environment variables that are EESSI version specific
+  if convertToCanonical(eessi_version) > convertToCanonical("2023.06") then
+    setenv ("EASYBUILD_PREFER_PYTHON_SEARCH_PATH", "EBPYTHONPREFIXES")
```
boegel, line 223: I don't have strong feelings about this, and I do see the appeal to use `$EBPYTHONPREFIXES` rather than `$PYTHONPATH` (works better when using Python virtual environments), but this is in some sense also uncharted territory, so we're bound to run into surprises/problems because we're using this.

Also, not relying on the standard mechanism (`$PYTHONPATH`) is no doubt going to confuse end users at some point.

So, we should clearly motivate why we're going down this road, since we can't change our mind in the middle, this is an all-or-nothing choice...

EESSI-extend-easybuild.eb
```diff
@@ -196,10 +201,30 @@ elseif (project_modulepath ~= nil) then
...
+    setenv ("EASYBUILD_MODULE_SEARCH_PATH_HEADERS", "include_paths")
+    setenv ("EASYBUILD_SEARCH_PATH_CPP_HEADERS", "include_paths")
```
boegel, line 225: Somewhat similar, why don't we just stick to the default (using `$CPATH`) here, is the case strong enough to go non-standard here by using `$C_INCLUDE_PATH` & co instead of `$CPATH`?

boegel (changes requested):

EESSI-extend-easybuild.eb
ocaisa: The motivation is exactly that, better `venv` behaviour, but I agree with the main concern about how hard it is to roll back. I was actually hoping you would be the one to sooth me.

End users can still use PYTHONPATH, but we wouldn't. Or are you more concerned about those using `EESSI-extend`?

EESSI-extend-easybuild.eb
```diff
@@ -196,10 +201,30 @@ elseif (project_modulepath ~= nil) then
...
     load(latest("EasyBuild"))
 end
+easybuild_version = os.getenv("EBVERSIONEASYBUILD") or easybuild_version
+eessi_version = os.getenv("EESSI_VERSION") or "2023.06"
+
+-- Set environment variables that are EasyBuild version specific
+if convertToCanonical(easybuild_version) > convertToCanonical("4") then
+  setenv ("EASYBUILD_STRICT_RPATH_SANITY_CHECK", "1")
+  setenv ("EASYBUILD_CUDA_SANITY_CHECK_ERROR_ON_FAILED_CHECKS", "1")
+  setenv ("EASYBUILD_ENV_FOR_SHEBANG", "1")
```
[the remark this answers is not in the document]
ocaisa: Ok, reverting this one

EESSI-extend-easybuild.eb
ocaisa: The issue and resolution is in https://github.com/easybuilders/easybuild-framework/issues/3331#issuecomment-2694517824 , but that doesn't really help me to be any clearer on what we should do. 

@Flamefire You were involved in that discussion, what do you think? Is it enough to only change `EASYBUILD_MODULE_SEARCH_PATH_HEADERS`?

Flamefire: I have seen issues fixed with the new `EASYBUILD_MODULE_SEARCH_PATH_HEADERS` but am not aware of the same for `EASYBUILD_SEARCH_PATH_CPP_HEADERS` but it seems sensible and I don't expect issues

I'd say this one is less problematic. If stuff really does break we can unset/change `EASYBUILD_SEARCH_PATH_CPP_HEADERS` because it isn't ingrained in module files except if the build did pick up wrong includes but I don't see how

EESSI-extend-easybuild.eb
boegel: I'm mostly concerned about people being confused as to how Python packages can be imported from non-standard locations while we're not using `$PYTHONPATH`.

The `$EBPYTHONPREFIXES` approach seems robust, but it's very custom.

That said, I'm not opposing it, just making sure we're aware that we're (yet again) taking a non-standard route here, and there may be dragons along that route...

We should probably add some documentation on the different ways in which we divert from default EasyBuild configuration, and explain the motivation behind those settings => https://github.com/EESSI/docs/issues/519

Flamefire: `EBPYTHONPREFIXES` isn't the default? I thought we would change that for EB 5

EESSI-extend-easybuild.eb
boegel: Yeah, that argument works in both directions

ocaisa: I don't think it completely does go both ways. The problem is that `CPATH` is set in module files and that takes precedence, we can always increase precedence _in our builds_ by setting `CPATH` in a build (to the value of `C_INCLUDE_PATH`) if we need to. We could indeed also handle the reverse.

The issue is more on the user side, you are asking them to diagnose and do the same thing if _they_ run into the `CPATH` problem, whereas they won't see it in the other case.

ocaisa: (see https://github.com/easybuilders/easybuild-framework/pull/4655/files)

casparvl: Ah, so @Flamefire 's earlier comment that this is more easily revertable

> If stuff really does break we can unset/change EASYBUILD_SEARCH_PATH_CPP_HEADERS because it isn't ingrained in module files except if the build did pick up wrong includes but I don't see how

applies to `EASYBUILD_SEARCH_PATH_CPP_HEADERS`, but changing `EASYBUILD_MODULE_SEARCH_PATH_HEADERS` is not (as easily) revertible, right?

boegel: Same here, worst case we can actually back out later by patching or re-generating the module files, but it likely won't come to that.

To let's move forward by configuring EasyBuild so it doesn't fiddle with `$CPATH`

PR #27, opened 2025-07-03
Make `EESSI-extend` support accelerator installations

ocaisa:

PR #46, opened 2025-07-30
Don't use an `accel` subdirectory when installing CUDA and libraries to unbreak symlinks

ocaisa:
Currently, the script that installs CUDA and cu* libraries to unbreak symlinks has different behaviour depending on whether or not a GPU is available on the node where it is run (due to https://github.com/EESSI/software-layer-scripts/pull/27 where `accel` subdirectories are enforced for site installations).

This change standardises behaviour, ignoring the subdirectory, as this is what the hook that creates the symlinks expects (and is ok since the packages being installed are binary installations)


casparvl (approved):
Fixes the issue I had on a bot instance I tested here https://github.com/SURF-hpcv/software-layer/pull/1#issuecomment-3135659835 

Tested, and found to work here: https://github.com/SURF-hpcv/software-layer-scripts/pull/2#issuecomment-3140290432

PR #48, opened 2025-07-31
re-enable failing accel-related CI checks for 2025.06

bedroge:
We left these disabled in #40, because they're currently failing due to missing `accel` dirs. This PR is just here to make sure that they get enabled when we've ingested some accel software (or when we find another way to prevent them from failing).


ocaisa (approved):
LGTM

PR #50, opened 2025-08-01
Add functionality to `eessi_container.sh` for specifying if a CVMFS repo should be bind or fuse mounted

bedroge:
Just discussed this a bit with @ocaisa on Slack, and had some code from an old branch that could more or less do this already, so I opened a PR for it.

Example:
```
./eessi_container.sh -v -r software.eessi.io,access=rw,mount=bind
```
This will bind mount the repo from the host into the container.

Probably needs to be tested a bit more, and could add some tests to CI.


eessi_container.sh
```diff
@@ -756,14 +756,40 @@ for cvmfs_repo in "${REPOSITORIES[@]}"
...
+                exit ${REPOSITORY_ERROR_EXITCODE}
+            fi
+        fi
+    done
+    # if a mount mode was not specified, we use a bind mount if the repository is available on the host,
+    # and otherwise we use a fuse mount
+    if [[ -z ${cvmfs_repo_mount} ]]; then
+        cvmfs_repo_mount="fuse"
+        if [[ -x $(command -v cvmfs_config) ]] && cvmfs_config probe ${cvmfs_repo_name} >& /dev/null; then
+            cvmfs_repo_mount="bind"
```
boegel, line 781: @bedroge Hmm, not sure I like this, since it changes the default behavior.

I think it's better to only to bind mount if explicitly requested, to retain same (default) behavior across different systems.

Auto-enabling the bind mount when the CernVM-FS repo is mounted leaves us vulnerable in case of trouble with that mount (like an outdated sync, CernVM-FS cache trouble, etc.)

bedroge: Changed in https://github.com/EESSI/software-layer-scripts/pull/57.

PR #54, opened 2025-08-05
fix typo in `EESSI-extend` + install CUDA in host injections earlier in the software install scripts (+ deploy `EESSI-extend/2025.06`)

ocaisa:
#45 Was not actually deployed

### DIFF
--- a/EESSI-extend-easybuild.eb
+++ b/EESSI-extend-easybuild.eb
@@ -70,7 +70,7 @@ modextravars = {
 # EASYBUILD_INSTALLPATH=${EESSI_PREFIX}/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR}
 # EASYBUILD_SOURCEPATH=${WORKDIR}/easybuild/sources:${EESSI_SOURCEPATH}
 #
-# And also some optional ones based on the kind of installation
+# And also some optional ones based on the installation mode
 # EASYBUILD_SET_GID_BIT
 # EASYBUILD_GROUP_WRITABLE_INSTALLDIR
 # EASYBUILD_UMASK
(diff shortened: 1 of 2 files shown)


bedroge:
Looks like it's using an old version of EB for this installation:
```
  1) EasyBuild/5.1.1 => EasyBuild/4.9.4
```

boegel:
> Unfortunately we cannot build for a64fx, this will then be dangerously out of sync!

How are we going to deal with this?

For this particular "build", we don't actually need access to an A64FX system at all.

bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-jsc architecture:aarch64/nvidia/grace

boegel:
> In case you want to try: here's the updated config [EESSI/bot-configs#39](https://github.com/EESSI/bot-configs/pull/39) (untested, so you may hit some issues if I forgot to update some item)

We also need to update the bot instance when, which implies that we need a new bot release.
But perhaps https://github.com/EESSI/eessi-bot-software-layer/issues/330 should be fixed first?

boegel:
bot: status instance:eessi-bot-deucalion

casparvl (approved):

boegel:
@Neves-P This PR contains tarballs for both `2023.06` and `2025.06` EESSI versions.

I'm guessing those should be handled through separate `EESSI/staging_bundles` PRs?

For now, I only see a staging PR for the `2023.06` tarballs...
Are we monitoring both buckets?

boegel:
> The error being reported is
> 
> ```
> ERROR: Failed to parse configuration options: 'Found 3 environment variable(s) that are prefixed with EASYBUILD but do not match valid option(s): EASYBUILD_CUDA_SANITY_CHECK_ERROR_ON_FAILED_CHECKS,EASYBUILD_FAIL_ON_MOD_FILES_GCCCORE,EASYBUILD_STRICT_RPATH_SANITY_CHECK'
> ```
> 
> now to find out why...
> 
> EDIT: A CUDA check somewhere is reverting to EB 4.9.4 which is causing the problem. We can sidestep it by unloading the module after a successful install

Proper fix for this is in #66

Issue #56, opened 2025-08-05
Unable to change host-injections when bind-mounting CVMFS from a host that has non-default host-injections location


casparvl:
In [my build](https://github.com/EESSI/software-layer/pull/1147#issuecomment-3155546690) I'm getting a lot of 

```
mktemp: failed to create file via template '/cvmfs/software.eessi.io/versions/2023.06/init/eessi_defaults.XXXXXX': Permission denied
/gpfs/work1/1/eessibot/eessi-bot-surf/jobs/2025.08/pr_1147/event_7c214fd0-720b-11f0-84af-0ecb387b6cc4/run_000/linux_x86_64_amd_zen4/eessi.io-2023.06-software/software-layer-scripts/install_scripts.sh: line 51: : No such file or directory
sed command failed
mktemp: failed to create file via template '/cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash.XXXXXX': Permission denied
/gpfs/work1/1/eessibot/eessi-bot-surf/jobs/2025.08/pr_1147/event_7c214fd0-720b-11f0-84af-0ecb387b6cc4/run_000/linux_x86_64_amd_zen4/eessi.io-2023.06-software/software-layer-scripts/install_scripts.sh: line 51: : No such file or directory
sed command failed
mktemp: failed to create file via template '/cvmfs/software.eessi.io/versions/2023.06/init/lmod/csh.XXXXXX': Permission denied
/gpfs/work1/1/eessibot/eessi-bot-surf/jobs/2025.08/pr_1147/event_7c214fd0-720b-11f0-84af-0ecb387b6cc4/run_000/linux_x86_64_amd_zen4/eessi.io-2023.06-software/software-layer-scripts/install_scripts.sh: line 51: : No such file or directory
sed command failed
mktemp: failed to create file via template '/cvmfs/software.eessi.io/versions/2023.06/init/lmod/fish.XXXXXX': Permission denied
/gpfs/work1/1/eessibot/eessi-bot-surf/jobs/2025.08/pr_1147/event_7c214fd0-720b-11f0-84af-0ecb387b6cc4/run_000/linux_x86_64_amd_zen4/eessi.io-2023.06-software/software-layer-scripts/install_scripts.sh: line 51: : No such file or directory
sed command failed
mktemp: failed to create file via template '/cvmfs/software.eessi.io/versions/2023.06/init/lmod/ksh.XXXXXX': Permission denied
[...]

PR #57, opened 2025-08-05
Always use fuse mounts by default in EESSI container, only use bind mounts when explicitly requested

bedroge:
Solves https://github.com/EESSI/software-layer-scripts/issues/56 by reverting the change which made bind mounts the default on systems where `/cvmfs/software.eessi.io` mounted on the host. This could cause issues with the host injections folder, if this system uses a non-standard path (something else than `/opt/eessi`).

Bind mounts can still be used, but have to be explicitly enabled by using something like:
`./eessi_container.sh -r myrepo,mount=bind`

Also improves the parsing of some variables, which could have a trailing newline in case they are the last/only argument. It's removing them using `${myvar%$'\n'}` (see https://unix.stackexchange.com/a/57128).

### DIFF
--- a/eessi_container.sh
+++ b/eessi_container.sh
@@ -762,21 +762,28 @@ do
     [[ ${VERBOSE} -eq 1 ]] && echo "add fusemount options for CVMFS repo '${cvmfs_repo}'"
     # split into name, access mode, and mount mode
     readarray -td, cvmfs_repo_args <<<"$cvmfs_repo"
-    cvmfs_repo_name=$(sed -e 's/\\n//g' <<< "${cvmfs_repo_args[0]}")
+    cvmfs_repo_name=${cvmfs_repo_args[0]%$'\n'} # remove possible trailing newline
     cvmfs_repo_access="${ACCESS}" # initialize to the default access mode
+    cvmfs_repo_mount="fuse" # use fuse mounts by default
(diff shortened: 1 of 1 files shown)


bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-jsc architecture:aarch64/nvidia/grace

bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-jsc architecture:aarch64/nvidia/grace

bedroge:
I've made some additional changes in https://github.com/EESSI/software-layer-scripts/pull/57/commits/6610e7e22257061115c05ca3d3c35e410617c151, as I just found other parts in the scripts where we loop over the repositories and where it only checked for `,access=*` (while there could be `mount=*` now as well, in any order).

bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-jsc architecture:aarch64/nvidia/grace

casparvl (approved):
I think we should _also_ investigate if we can somehow do bind-mounting on hosts that have the repo mounted _and_ have a non-standard `host-injections`, this solves some immediate build issues. Looks good to me, test by Bob was succesful.

I've removed the `cowsay` test easystack, hope that's ok with you @bedroge , but then it can be merged right away :)

PR #63, opened 2025-08-14
Fix upload for multiple GPU targets

casparvl:
This fixes the issue reported here https://github.com/EESSI/eessi-bot-software-layer/issues/333 which didn't turn out to be a bot issue, but a naming issue: if the names (excluding timestamp) are the same, the bot will only upload the latest (if the upload policy is 'latest'). So, we need to add the GPU architecture in the name, so that if we build e.g. for `zen2`+`cc70` and `zen2` + `cc80`, that _both_ are uploaded (and not just the one that was built the latest).

Proof that this PR fixes the issue is in https://github.com/casparvl/software-layer-scripts/pull/1 , where I've first reproduced the issue (i.e. you see only one of the two tarballs being deployed). Then, with the fix implemented, I did a rebuild and redeploy, and it deployed _both_ tarballs.


Neves-P (approved):
lgtm!

PR #78, opened 2025-08-31
Initial notification to user that Terms of Use exist

ocaisa:
This is a first pass solution, we can do something more fancy in follow-up PRs

### DIFF
--- a/init/bash
+++ b/init/bash
@@ -48,5 +48,6 @@ if [ $? -eq 0 ]; then
     #show_msg ""
 
     echo "Environment set up to use EESSI (${EESSI_VERSION}), have fun!"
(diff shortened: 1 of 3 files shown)


init/bash
```diff
@@ -48,5 +48,6 @@ if [ $? -eq 0 ]; then
     #show_msg ""
 
     echo "Environment set up to use EESSI (${EESSI_VERSION}), have fun!"
+    echo -e "\033[1;33mPlease note that by configuring and using the EESSI software stack,\nyou agree to abide by the EESSI Terms of Use\n(see https://www.eessi.io/docs/governance/terms_of_use/).\033[0m"
```
Crivella, line 51: Should we check if the terminal supports ANSI escape codes?

ocaisa: @Crivella I couldn't get colours with Lmod anyway, so I've reverted to using `*****` to get attention

PR #80, opened 2025-09-03
pick up RISC-V software installations from `/cvmfs/dev.eessi.io/riscv` for EESSI 2025.06

bedroge:
Version 2023.06 still redirects to `/cvmfs/riscv.eessi.io` for RISC-V clients, but for 2025.06 this will make the module use the init scripts and compat layer from `software.eessi.io` and the software installations from `dev.eessi.io/riscv`. In order to do that, I've added a `eessi_archdetect_prefix` variable which is set early, to ensure that it will look for archdetect in the production repo (instead of the dev repo). If the dev repo is not available on the host, it will print an error.

### DIFF
--- a/init/modules/EESSI/2023.06.lua
+++ b/init/modules/EESSI/2023.06.lua
@@ -22,15 +22,19 @@ if (subprocess("uname -m"):gsub("\n$","") == "riscv64") then
         eessi_version = os.getenv("EESSI_VERSION_OVERRIDE") or "20240402"
         eessi_repo = "/cvmfs/riscv.eessi.io"
         eessi_prefix = pathJoin(eessi_repo, "versions", eessi_version)
-        LmodMessage("RISC-V architecture detected, but there is no RISC-V support yet in the production repository.\n" ..
-                    "Automatically switching to version " .. eessi_version .. " of the RISC-V development repository " .. eessi_repo .. ".\n" ..
-                    "For more details about this repository, see https://www.eessi.io/docs/repositories/riscv.eessi.io/.")
+        if mode() == "load" then
+            LmodMessage("RISC-V architecture detected, but there is no RISC-V support yet in the production repository.\n" ..
+                        "Automatically switching to version " .. eessi_version .. " of the RISC-V development repository " .. eessi_repo .. ".\n" ..
+                        "For more details about this repository, see https://www.eessi.io/docs/repositories/riscv.eessi.io/.")
+        end
     elseif (eessi_version == "2025.06") then
         eessi_repo = "/cvmfs/dev.eessi.io/riscv"
         eessi_version = os.getenv("EESSI_VERSION_OVERRIDE") or eessi_version
         eessi_prefix = pathJoin(eessi_repo, "versions", eessi_version)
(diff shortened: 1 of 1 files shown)


init/modules/EESSI/2023.06.lua
```diff
@@ -13,12 +13,24 @@ conflict("EESSI")
...
+        eessi_repo = "/cvmfs/dev.eessi.io/riscv"
+        LmodMessage("This EESSI production version only provides a RISC-V compatibility layer,\n" ..
+                    "software installations are provided by the EESSI development repository at " .. eessi_repo .. ".\n")
+        if not isDir(eessi_repo) then
+            LmodError("The EESSI development repository dev.eessi.io is not mounted on your system.\n" ..
+                      "This is required for RISC-V systems.")
+        end
+    end
 end
 local eessi_prefix = pathJoin(eessi_repo, "versions", eessi_version)
```
boegel, line 35: For consistency, this should be moved up and used to define `eessi_archdetect_prefix`?

init/modules/EESSI/2023.06.lua
```diff
@@ -13,12 +13,24 @@ conflict("EESSI")
...
-                "Automatically switching to version " .. eessi_version .. " of the RISC-V development repository " .. eessi_repo .. ".\n" ..
-                "For more details about this repository, see https://www.eessi.io/docs/repositories/riscv.eessi.io/.")
+    if (eessi_version == "2023.06") then
+        eessi_version = os.getenv("EESSI_VERSION_OVERRIDE") or "20240402"
+        eessi_repo = "/cvmfs/riscv.eessi.io"
+        LmodMessage("RISC-V architecture detected, but there is no RISC-V support yet in the production repository.\n" ..
+                    "Automatically switching to version " .. eessi_version .. " of the RISC-V development repository " .. eessi_repo .. ".\n" ..
+                    "For more details about this repository, see https://www.eessi.io/docs/repositories/riscv.eessi.io/.")
+    elseif (eessi_version == "2025.06") then
+        eessi_repo = "/cvmfs/dev.eessi.io/riscv"
```
boegel, line 26: Should we also pick up on `$EESSI_VERSION_OVERRIDE` here, something like:

```suggestion
        eessi_repo = "/cvmfs/dev.eessi.io/riscv"
        eessi_version = os.getenv("EESSI_VERSION_OVERRIDE") or eessi_version
```

I imagine for RISC-V we'll have "iterations" in `dev.eessi.io`, like `2025.06-202509` or something

boegel (changes requested):

init/modules/EESSI/2023.06.lua
bedroge: Done in https://github.com/EESSI/software-layer-scripts/pull/80/commits/001cea1b68c4a815f8dc7c17899f80e22d8b50e7. It does mean that we have to override that variable as well for the different RISC-V versions, but I've added that too in the same commit.

ocaisa (approved):
LGTM, @bedroge can you just confirm that it works as expected

EDIT: you already have above

ocaisa:
Deployment PR merged

PR #82, opened 2025-09-08
Don't print messages from hook in `--terse` mode

Flamefire:
The output from the hook such as "Using custom configure options" makes the output not machine-readable which it should be in `--terse` mode.

Check the option in a hook-function before calling the original one


ocaisa:
@Flamefire Can you sync this now and I will get it deployed

PR #95, opened 2025-09-29
add pre-configure hook for GRASS that replaces paths of filtered dependencies

TopRichard:


eb_hooks.py
```diff
@@ -409,6 +409,26 @@ def parse_hook_cgal_toolchainopts_precise(ec, eprefix):
         raise EasyBuildError("CGAL-specific hook triggered for non-CGAL easyconfig?!")
 
 
+def parse_hook_grass(ec, *args, **kwargs):
```
bedroge, line 412: I would make this a `pre_configure_hook` instead. Also, it may be better/safer to still use the the configuration options, but make sure that they get the correct path to the libraries/header filers of the compat layer.  This should ensure that it picks up those libraries/headers, and not accidentally would pick up things from e.g. the build container's OS. Replacing them is not entirely straightforward (`self.cfg.update` would add them instead of replace them), but can be done with a regex. Looks like this should work:

```
def pre_configure_hook_grass(self, *args, **kwargs):
    """
    Pre-configure hook for GRASS to remove filtered deps specific configopts lines for readline, zlib, and bzlib
    """
    if self.name == 'GRASS':
        # determine path to Prefix installation in compat layer via $EPREFIX
        eprefix = get_eessi_envvar('EPREFIX')

        # Dependencies for which the configuration options need to be replaced
        filtered_deps = ['readline', 'zlib', 'bzlib']
        for dep in filtered_deps:
            self.cfg['configopts'] = re.sub(fr'--with-{dep}-includes=\S*', f'--with-{dep}-includes=$EPREFIX/usr/include', self.cfg['configopts'])
            self.cfg['configopts'] = re.sub(fr'--with-{dep}-libs=\S*', f'--with-{dep}-libs=$EPREFIX/usr/lib64', self.cfg['configopts'])
        print_msg("Using custom configure options for %s", self.name)
    else:
        raise EasyBuildError("GRASS-specific hook triggered for non-GRASS easyconfig?!")
```

TopRichard: Replaced the hook with a ` pre_configure_hook`

bedroge:

PR #102, opened 2025-10-05
fix script to init EESSI for `csh` shells, fixes also the CI for `csh` and speeds up CI

trz42:
The init script for `csh`, e.g., `/cvmfs/software.eessi.io/versions/2023.06/init/lmod/csh`, wasn't working properly (some syntax issues with oneliners for `if-then-endif` constructs as well as slightly off syntax for running commands (`csh` uses backticks and does not support `$(...)` syntax).

CI for `csh` was fixed, mostly by adjusting to different syntax required and working around that - in non-interactive shells - aliases (such as `module`) aren't retained for subsequent commands after sourcing the init script. Some polishing was applied to the test script (`test_init_scripts.sh`) as well (mostly replacing tabs with spaces).

Finally, installation of CernVM-FS (and other packages) was sped up by preventing the processing of the `man-db` trigger. Unless we need `man` pages in CI this should be fine. Note, only for the workflow `tests_init_module.yml`.

### DIFF
--- a/.github/workflows/tests_init_module.yml
+++ b/.github/workflows/tests_init_module.yml
@@ -59,8 +59,5 @@ jobs:
         run: |
           export EESSI_VERSION=${{matrix.EESSI_VERSION}}
           export EXPECTED_EASYBUILD_VERSION=${{matrix.EXPECTED_EASYBUILD_VERSION}}
-          echo "==============================================="
-          cat .github/workflows/scripts/test_init_scripts.sh
-          echo "==============================================="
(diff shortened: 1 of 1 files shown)


.github/workflows/tests_init_module.yml
```diff
@@ -27,6 +27,12 @@ jobs:
       - name: Check out software-layer repository
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
+      - name: Avoid processing man-db trigger
```
ocaisa, line 30: This is good, I'm going to steal this for the EESSI GitHub Action

PR #114, opened 2025-10-23
Fix issues for `LD_LIBRARY_PATH` via `EESSI` module

ocaisa:
Compat layer binaries are not rpathed, so if host locations like `/usr/lib` are in `LD_LIBRARY_PATH` things will break. Fix this in the `EESSI` module.

See https://gitlab.com/eessi/support/-/issues/198 for a lot of context.


ocaisa:

init/modules/EESSI/2023.06.lua
```diff
@@ -210,6 +210,22 @@ if os.getenv("EESSI_MODULE_STICKY") then
...
+append_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", os.getenv("LD_LIBRARY_PATH") or "")
+-- on unload the variable will no longer exist
+if mode() == "load" then
+    -- remove any standard paths that can interfere with the compat layer
+    remove_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", "/lib")
+    remove_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", "/lib64")
+    remove_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", "/usr/lib")
+    remove_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", "/usr/lib64")
+    remove_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", "/usr/local/lib")
+    remove_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", "/usr/local/lib64")
```
casparvl: I'm confused, didn't you intend to remove from `LD_LIBRARY_PATH` here? I.e. now, you are essentially leaving `LD_LIBRARY_PATH` unsanitized, while you're cleaning `EESSI_DEFAULT_HOST_LD_LIBRARY_PATH` (which I think you meant to keep as a the original value so you can restore later)

casparvl: I mean, I see that your test passes, I just don't understand how... I don't think it should pass with the current code...

init/modules/EESSI/2023.06.lua
```diff
@@ -210,6 +210,22 @@ if os.getenv("EESSI_MODULE_STICKY") then
     load_message = load_message .. " (requires '--force' option to unload or purge)"
 end
 
+-- Filter system paths from LD_LIBRARY_PATH
+-- Needs to be reversible so first make a copy
+append_path ("EESSI_DEFAULT_HOST_LD_LIBRARY_PATH", os.getenv("LD_LIBRARY_PATH") or "")
```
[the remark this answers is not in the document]
ocaisa: Would it help to do more extensive commenting?

init/modules/EESSI/2023.06.lua
ocaisa: Hmm, these paths are not quite enough (looking at https://gitlab.com/eessi/support/-/issues/198#note_2780502960), we need to add directories like `/usr/lib/x86_64-linux-gnu`, maybe I need a catch-all here

ocaisa:
Argh, this can't work because Lmod _itself_ that is broken by `LD_LIBRARY_PATH` so we never get to the point where the module can solve the problem.

ocaisa:
You can fix Lmod with
```
export LMOD_CMD="env -u LD_LIBRARY_PATH $LMOD_CMD"
```

ocaisa:
This is brittle. While the idea is ok, it only works if Lmod can still function. This might not be the case if you have a system Lmod and the EESSI Lmod activated. I will fix it for the case where it does work, but then it has limited application (but the breaking case would be rare enough: you need EESSI Lmod _and_ locally installed lua installed in a location that appears in `LD_LIBRARY_PATH`).

ocaisa:
The specific case that triggered this PR was in RStudio, but that probably has a better solution https://github.com/rstudio/rstudio/pull/17390

PR #125, opened 2025-11-12
use `head -n 1` instead of `head -1` (likewise for `tail`)

boegel:
On some systems (like MareNostrum 5) `head -1` and `tail -1` doesn't work:

```
$ echo boo | tail -1
tail: option used in invalid context -- 1
```

Seems to be caused by an alias that's used:
```
alias tail='tail ---disable-inotify'
```

On MareNostrum 5 GPU nodes, this leads to broken GPU detection:
```
tail: option used in invalid context -- 1
ERROR: accelerator detection with archdetect worked, but no result was returned?!
```


ocaisa (approved):
LGTM

PR #132, opened 2025-11-18
Remove workaround/hook for GROMACS SVE issue

bedroge:
First testing this for GROMACS 2024.1 and Neoverse V1 + Grace, but the hook was also applied for 2024.3+4 on Grace CPUs. So, if this works, I'll do test builds for the latter versions as well, as I'm not sure if those versions had the same issues as encountered on for instance A64FX (if so, they should be solved as well with the new easyconfig).

### DIFF
--- a/EESSI-extend-easybuild.eb
+++ b/EESSI-extend-easybuild.eb
@@ -96,11 +96,23 @@ sysroot = os.getenv("EESSI_EPREFIX")
 -- Check if we have GPU capabilities and configure CUDA compute capabilities
 eessi_accelerator_target = os.getenv("EESSI_ACCELERATOR_TARGET")
 if (eessi_accelerator_target ~= nil) then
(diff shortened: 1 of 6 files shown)


bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws for:arch=aarch64/neoverse_v1
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-jsc for:arch=aarch64/nvidia/grace

ocaisa (approved):
LGTM

bedroge:
> I guess this was accidentally built & deploy twice for EESSI version 2023.06, but no harm done since the `eb_hooks.py` was identical in both?

It was built twice because I wanted to do test builds with GROMACS for both Neoverse V1 and Grace. To prevent a tarball with GROMACS from getting deployed, I did the final builds for both targets as well, otherwise I had to close the PR and make a new one. We could have set one tarball to `REJECTED` in the staging PR, but there's indeed no harm in deploying the exact same thing twice.

PR #135, opened 2025-11-24
take software layer version suffix into account when creating a tarball

bedroge:
When we build for a version with a suffix (e.g. `2025.06-001`), the tarball step should actually check the directory name with the suffix in the overlay's `upper` dir. Note that it already sources `eessi_defaults` on line 293, this should set `$EESSI_SOFTWARE_LAYER_VERSION_SUFFIX`.


bedroge:
Let's make sure that it doesn't break anything for regular builds.

bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws for:arch=x86_64/intel/haswell

boegel (approved):

PR #115, opened 2025-10-24
Enhance LLVM hook to cover changes in easyblocks #3881

Crivella:
This should make the hook work again after `$EBROOTXXX` is not what is being checked anymore directly in the Easyblock after

- https://github.com/easybuilders/easybuild-easyblocks/pull/3881

This should always work as we are enforcing not doing any configure specific action in the `__init__` to not break things with bundles and we are still checking `$EBROOTXXX` to get the actual paths to the libraries (just not to see if they are a dependency)

https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/l/llvm.py#L531-L533

PR #136, opened 2025-12-05
Allow LLVM hook to run from within a Bundle

Crivella:
Fix on top of

- #115 

To also allow the new fix to work when `EB_LLVM` is called from within a bundle

If we do not want this generic way of going through the components of the Bundle we will need to split the hook for `LLVM` and `ROCm-LLVM` (maybe reusing a common function under the hood)


casparvl:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-deucalion for:arch=aarch64/a64fx
bot: build repo:eessi.io-2025.06-software instance:eessi-bot-deucalion for:arch=aarch64/a64fx

PR #141, opened 2025-12-19
check if `ld` wrappers are actually available in `post_prepare_hook_gcc_prefixed_ld_rpath_wrapper`

bedroge:
On RISC-V a build of GCC failed (https://github.com/EESSI/dev.eessi.io-riscv/pull/3) because `ld.gold` is not available (not in the compat layer and neither in the build container), leading to:
```
  File "/cvmfs/software.eessi.io/versions/2025.06/init/easybuild/eb_hooks.py", line 364, in post_prepare_hook
    POST_PREPARE_HOOKS[self.name](self, *args, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/cvmfs/software.eessi.io/versions/2025.06/init/easybuild/eb_hooks.py", line 335, in post_prepare_hook_gcc_prefixed_ld_rpath_wrapper
    wrapper_dir = os.path.dirname(wrapper)
  File "<frozen posixpath>", line 178, in dirname
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

This PR adds a check that will skip the wrapper if it's not available.

My test build with this updated hook is still running on RISC-V, but it already got past the prepare step, so it seems to work fine.

### DIFF
--- a/EESSI-extend-easybuild.eb
+++ b/EESSI-extend-easybuild.eb
@@ -45,13 +45,13 @@ description = """
 toolchain = SYSTEM
 
 # All the dependencies we filter in EESSI
-local_deps_to_filter = "binutils,bzip2,DBus,flex,gettext,gperf,help2man,intltool,libreadline,makeinfo,ncurses,ParMETIS,util-linux,XZ,zlib"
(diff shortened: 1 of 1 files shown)


bedroge:
Ran into the issue described here: https://github.com/EESSI/software-layer/issues/552#issuecomment-2073044401. Trying the fix from https://github.com/EESSI/software-layer-scripts/pull/141/commits/cc39efb12efe458538e1486a9cea3924fa750a97 now.

PR #146, opened 2025-12-24
Use module-only when a CUDA Compute Capability is requested that is incompatible with the CUDA toolkit version used

casparvl:
Add some initial changes to the hooks to make sure to install with --module-only if this is CUDA-12.6 based but targets CC100 or CC120.

This still needs to be completed. Also, I could potentially make it more clever and match anything that's >=CC100.

Edit 12-01: This PR now does a four things.

1. Make the way in which unsupported modules are handled more generic. With that, if we ever need to add new cases of unsupported module, the thing we need to do is add the new case to `is_unsupported_module`
2. Use that generic approach to check for compatibility between CUDA Compute Capability and CUDA toolkit version - and treat it as an unsupported module if it's not compatible
3. Replace `--cuda-compute-capabilities=9.0a` with `--cuda-compute-capabilities=9.0` for `cuDNN-9.5.0.50` and similarly strip the suffix for `10.0f` and `12.0f` for `cuDNN-9.10.1.4`.

To achieve this, I first made the current mechanism that was there for handling the Zen4+foss-2022b incompatibility more generic. 

- Replace specific hooks like `parse_hook_zen4_module_only` (took care of adding the LmodError in a `modluafooter`), `pre_prepare_hook_ignore_zen4_gcccore1220_error` (sets env var to suppress the LmodError when building other software on top, so that you create modules dependent on the unsupported module) and `post_prepare_hook_ignore_zen4_gcccore1220_error` (unset that env var), with more generic hooks 
- Create a NamedTuple to hold two pieces of information:
  - The name of the environment variable to suppress the associated LmodError
  - The text for the LmodError
- Set that NamedTuple as an attribute
- Make other hooks use that attribute to set the relevant environment variable & error message
- Moved setting the luafooter to the `pre_module_hook`, since some information relevant to determining if a module is unsupported may not be available as early as the `parse_hook` (such as the requested CUDA compute capability) 

Then, I 

- implemented logic to check for the CUDA compatibility
- implemented a message & env-var for the CUDA case in `is_unsupported_module`.

I've left some reviewer-comments in the changed files to make it easier for anyone reviewing this to see _why_ things were changed.

I then ran tests to a) validate that it still worked for zen4+foss-2022b and b) did what it should do for the CUDA CC vs Toolkit version (in)compatibility. Results are in summary below.

<details>

<summary>zen4+foss-2022b test</summary>

Environment:

```
module load EESSI/2023.06
module load EESSI-extend/2023.06-easybuild
```

Build log:

```
[...]

PR #147, opened 2026-01-13
Improve software-layer-script workflow

casparvl:
Changes needed to make builds use everything from the cloned software-layer-scripts, instead of from the deployed stuff in software.eessi.io.

This is a starting point for https://gitlab.com/eessi/support/-/issues/217.

Split off from https://github.com/EESSI/software-layer-scripts/pull/146/commits/24228048da52569255e86bb7fd0faa552f0cd4ad which was tested in e.g. https://github.com/EESSI/software-layer/pull/1351#issuecomment-3743386595

We may not want to merge this until we have sufficient checks (CI, bot) in place to ensure that tarballs can only be deployed if they were build from a merge commit of `software-layer-scripts`. That means we need:
- Software-layer CI to check if upstream `software-layer-scripts` commit was merged https://github.com/EESSI/software-layer/pull/1353
- Bot improvement to check if build was done using the latest `bot/commit_sha` (i.e. the same one as the CI check uses).

Edit 10-08-2026: actually, this is somewhat separate from the `software-layer` work to use a particular commit checksum. Making the build scripts use everything from the `software-layer-scripts` repository makes sense in any case: it prevents us from first having to deploy certain components before we're actually able to test them. The same applies to the modification in the order in which CI steps are executed, that's also to prevent us from first having to deploy `EESSI-extend-easybuild.eb` changes to production _before_ all the CI steps are run on those changes.


bedroge (commented):
Just to summarize how this used to work and how it will work after merging this PR:

Previously, an updated hooks file was copied to the CVMFS repo (with the overlay) before the builds were started (by `EESSI-install-software.sh`), and this was only done if a PR diff file showed that the hooks file was modified by the PR.

In the new situation it will just use the version from the git clone of software-layer-scripts; the build script clones a specific commit of that repo. Additional CI will make sure that software-layer PR tarballs can only be deployed if that commit belongs to a merged PR.

One thing that maybe requires some attention is possible situations where a software layer PR modifies that software-layer-scripts commit, does the builds, and then undoes those changes regarding the used commit. A reviewer may not immediately spot that the builds were done with a different commit (the "Files changed" tab won't show it, you would have to look at individual commits or notice that there was a commit after the builds were done), but maybe the CI or bot itself could also catch that?

EESSI-install-software.sh
```diff
@@ -425,6 +425,10 @@ else
             # load EasyBuild module (will be installed if it's not available yet)
             source ${TOPDIR}/load_easybuild_module.sh ${eb_version}
 
+            # Set EASYBUILD_HOOKS to use the hooks from the software-layer-scripts, to make it easier to use updated hooks
+            export EASYBUILD_HOOKS=$TOPDIR/eb_hooks.py
```
casparvl: We should make sure we can also pass the `$TOPDIR/eb_hooks.py` to `install_cuda_and_libraries.sh` as an optional argument. (since `install_cuda_and_libraries.sh` loads its own `EESSI-extend`, just setting an environment variable is not sufficient)

casparvl: Suggestion by Bob: instead, let's revise how `EESSI-extend` behaves: make it listen to an `EESSI_EASYBUILD_HOOKS_OVERRIDE` variable. That's also useful for sites in general if they want to extend our hooks.

This is also more robust, since any child process that would invoke `EESSI-extend` will have this environment variable set.

bedroge (approved):
Nice work!

PR #159, opened 2026-02-06
use Rust 1.91.1 instead of 1.88.0 for maturin 1.9.1 in 2025b toolchain

bedroge:


bedroge:
bot: build repo:eessi.io-2025.06-software instance:eessi-bot-deucalion for:arch=aarch64/a64fx

eb_hooks.py
```diff
@@ -632,6 +635,20 @@ def parse_hook_qt5_check_qtwebengine_disable(ec, eprefix):
         raise EasyBuildError("Qt5-specific hook triggered for non-Qt5 easyconfig?!")
 
 
+def parse_hook_rust_2025b(ec, eprefix):
+    """
+    Replace build dependency on Rust 1.88.0 (used by the 2025b toolchain) by 1.91.1,
```
boegel: I like being consistent, but this does mean we got in uncharted territory compared to EasyBuild...

Maybe we should only do this where we've seen it matter, i.e. for `maturin`?

bedroge: I didn't really like that it would then basically mix different Rust versions for e.g. Python-bundle-PyPI (which has a build dependency on Rust itself, but also on maturin). But I can give that a try and see if that works. There are not that many easyconfigs that depend on Rust, so we can easily extend that list later on.

bedroge:
Testing it locally with EESSI-extend shows that it still works fine after adding the (small) additional changes.
```
$ eb -M ~/easybuild/easybuild-easyconfigs/easybuild/easyconfigs/m/maturin/maturin-1.9.1-GCCcore-14.3.0.eb
== Temporary log file in case of crash /tmp/eb-tcwllz37/easybuild-j558tkrr.log
== found valid index for /home/bob/easybuild/cit-hpc-easybuild/easyconfigs, so using it...
== Running parse hook for maturin-1.9.1-GCCcore-14.3.0.eb...
== Replaced ('Rust', '1.88.0') build dependency by ('Rust', '1.91.1') for maturin 1.9.1

```

PR #163, opened 2026-02-17
Script for reproducing an existing stack for a new CPU target

casparvl:
This script is the result of joint development with @bedroge on implementing what was discussed in https://gitlab.com/eessi/support/-/issues/233

In summary: it generates easystack files that will allow you to replicate the software stack installed for a reference architecture. We will use this in practice to deploy software for `zen5`, which will also serve as a test for these scripts.

Known limitations (see https://github.com/EESSI/software-layer/issues/1430): doing the rebuild at the time of the original build (as will be done when generating easystacks with this script) will fail if the rebuild added _new dependencies_. To resolve this, we should do an iterative improvement of this script in which we replay rebuilds as well. This takes slightly more build time, but avoids these issues, thus resulting in a more seamless experience in reproducing the stack for a new target. This is listed as a TODO in https://github.com/EESSI/software-layer/issues/1430#issuecomment-4024929587


bedroge:
A very similar but slighly more complicated issue popped up for gst-plugins-bad in https://github.com/EESSI/software-layer/pull/1440. Again, it had to be solved by moving additional dependencies around in the easystacks, which was done in https://github.com/EESSI/software-layer/pull/1443. As this one was more complex, I'm beginning to lean towards changing the approach and handle rebuilds at the end.

But, for now, I think we should go ahead and merge this, as it has been used quite successfully for zen5. Issues that we encountered are listed in https://github.com/EESSI/software-layer/issues/1430. There are certainly things that can be improved in the script, but we can do that in follow-up PRs.

bedroge (approved):

PR #166, opened 2026-02-20
add hook that makes BLIS fall back to zen3 on zen5

bedroge:
This basically does the same thing as https://github.com/easybuilders/easybuild-easyblocks/pull/4034, but that doesn't work on our AWS Zen5 VM, becausae archspec detects it as zen4.


ocaisa (approved):
LGTM

PR #169, opened 2026-02-23
quick fix to hide the licenses directory

hvelab:
right now the builds are broken because the bot attempts to symlink everything and "licenses" dir exist in both software-layer and software-layer-scripts repo

the one here is outdated and deprecated


ocaisa (dismissed):
We can always bring it back from Git if need be

PR #176, opened 2026-03-09
add comment to EESSI-extend easyconfig file to note that `module-search-path-headers` and `search-path-cpp-headers` should not be set to same value in EasyBuild configuration

boegel:
follow-up for:
- #175
which we failed to deploy for all CPU targets...


boegel:
bot: build repo:eessi.io-2025.06-software instance:eessi-bot-aws-eu-south for:arch=x86_64/amd/zen5

PR #181, opened 2026-03-12
Set PyTorch parallelism limits for `aarch64/{generic,neoverse_n1,neoverse_v1}`, allow more failing tests for A64FX

bedroge:
Still keeping this as draft to potentially solve other issues.


EESSI-install-software.sh
```diff
@@ -456,6 +456,8 @@ else
             # load EasyBuild module (will be installed if it's not available yet)
             source ${TOPDIR}/load_easybuild_module.sh ${eb_version}
 
+            export EASYBUILD_HOOKS=$TOPDIR/eb_hooks.py
```
casparvl, line 459: Will need to be removed before merging this PR

PR #185, opened 2026-03-23
Directly check the location of `libcuda.so` where the linker expects to find it

### LINKED ISSUE #184: Hook for checking if CUDA driver is available does not work for `EESSI/2025.06`
When we updated the mechanism for adding CUDA drivers to 2025.06 we never updated the hook that injects a check for the drivers.

ocaisa:
Fixes #184


casparvl (approved):
Looks good to me, in that it generates an error when it should. One concern I have is that the error message may be a bit in accurate:

```
You can override this check by setting the environment variable EESSI_OVERRIDE_GPU_CHECK but the loaded application will not be able to execute on your system.
For more information on how to do this, see https://www.eessi.io/docs/site_specific_config/gpu/.
```
is in the wrong order. The "this" in 'more information on how to do THIS' _should_ apply to making sure EESSI can find the GPU drivers. In it's current place, it applies to overriding the check - that's not what we wanted to say.

I know it's a bit beside the point of this PR. I'll deploy it since we want this live ASAP and since you need to do so many builds to fix it. But if you agree that this is awkward, maybe you can fix it in a follow-up PR.

PR #193, opened 2026-03-26
Use sandbox mode which is more robust in singularity cleanup phase

julianmorillo:


trz42:
Could we use this optionally? It's been tested on RISC-V, but we haven't seen this on other systems. So, to not cause problems on the other systems, maybe the safest way forward would be to add an option and use that where needed.

julianmorillo:
@boegel 
[SingularityCE User Guide (runtime behavior changes)](https://docs.sylabs.io/guides/4.1/user-guide.pdf)
[SingularityCE 4.1 Admin/User notes (FUSE + sandbox fallback)](https://docs.sylabs.io/guides/4.1/admin-guide/new.html)

**The critical mechanism** here (**from** the **documentation** above) is that **SIF images require mounting (via FUSE)**. In the documentation above it can be read: "SIF/SquashFS container images will now be mounted with squashfuse… when kernel mounts are not available."
In an unprivileged HPC environment this becomes:
- **FUSE mount**
- Which requires **cleanup on exit** --> where errors in https://github.com/EESSI/dev.eessi.io-riscv/pull/35#issuecomment-4141628587 happen.

The documentation above also explains that **if mounting fails or is unsuitable --> Singularity uses a sandbox**: “If the FUSE mount fails, Singularity will fall back to extracting the container to a temporary sandbox.” and “fallback to extraction to a temporary sandbox directory”.
The key insight is that:
- **Sandbox = extracted directory**
- No mount needed
- No FUSE
- No teardown/cleanup complexity

bedroge:
> Could we use this optionally? It's been tested on RISC-V, but we haven't seen this on other systems. So, to not cause problems on the other systems, maybe the safest way forward would be to add an option and use that where needed.

This makes sense to me. The sandbox mode is probably (much) slower, especially when multiple jobs are extracting that image to a shared filesystem at the same time.

julianmorillo:
> > Could we use this optionally? It's been tested on RISC-V, but we haven't seen this on other systems. So, to not cause problems on the other systems, maybe the safest way forward would be to add an option and use that where needed.
> 
> This makes sense to me. The sandbox mode is probably (much) slower, especially when multiple jobs are extracting that image to a shared filesystem at the same time.

I already made it optional through an environment variable that should be defined in `site_config.sh`. Would that work?

julianmorillo:
Tested successfully in https://github.com/EESSI/dev.eessi.io-riscv/pull/103, and https://github.com/EESSI/dev.eessi.io-riscv/pull/105

bedroge (changes requested):
Could you move the eb_hooks.py changes to a separate PR, as they are unrelated?

The sandbox stuff looks good to me, though it would be good to also add an option to the script itself, as most of these things have a corresponding option. `-s` is already taken for `--save`, so we need another one (`-z`? or only use a long option `--sandbox`?)

julianmorillo:
Implemented all the suggested changes. I took the option of just using the long `--sandbox` option.

eessi_container.sh
```diff
@@ -118,6 +118,9 @@ display_help() {
   echo "                            [default: not set]; uses env var \$http_proxy if set"
   echo "  -y | --https-proxy URL  - provides URL for the env variable https_proxy"
   echo "                            [default: not set]; uses env var \$https_proxy if set"
+  echo "       --sandbox          - use sandbox mode (i.e. convert .sif image to sandbox and then run" 
+  echo "                            it instead"
+  echo "                            [default: not set]"
```
bedroge: Nitpicking a bit now, but while looking up how other commands order these options (in case you have a mixture of short and long ones), I realized that you can also have uppercase letter options.
So then I'd prefer `-S | --sandbox`, and you could move it below `--save` (ordered by short option, lowercase first).

eessi_container.sh
```diff
@@ -293,6 +296,10 @@ while [[ $# -gt 0 ]]; do
       export https_proxy=${HTTPS_PROXY}
       shift 2
       ;;
+    --sandbox)
+      SANDBOX=1
+      shift 1
+      ;;
```
bedroge: This would have to change accordingly, and can be moved up as well.

eessi_container.sh
```diff
@@ -1039,10 +1046,23 @@ for arg in "${PASS_THROUGH[@]}"; do
     ADDITIONAL_CONTAINER_OPTIONS+=(${arg})
 done
 
-echo "Launching container with command (next line):"
-echo "singularity ${RUN_QUIET} ${MODE} ${ADDITIONAL_CONTAINER_OPTIONS[@]} ${EESSI_FUSE_MOUNTS[@]} ${CONTAINER} $@"
-singularity ${RUN_QUIET} ${MODE} "${ADDITIONAL_CONTAINER_OPTIONS[@]}" "${EESSI_FUSE_MOUNTS[@]}" ${CONTAINER} "$@"
+# EESSI_SINGULARITY_SANDBOX is an environment variable (typically set in site_config.sh, if needed)
+if [[ -n "$EESSI_SINGULARITY_SANDBOX"  ||  $SANDBOX -eq 1 ]]; then
```
bedroge: Final tiny nitpick, to make the code style more consistent with the rest of the script: 
```suggestion
if [[ -n "${EESSI_SINGULARITY_SANDBOX}"  ||  ${SANDBOX} -eq 1 ]]; then
```

bedroge (changes requested):
Tested it, and works fine. Some final nitpicks and then we can merge this.

eessi_container.sh
julianmorillo: Done!

eessi_container.sh
julianmorillo: Done!

bedroge (approved):
Lgtm, thanks @julianmorillo!

Issue #201, opened 2026-04-14
NVIDIA GPU driver version written by link_nvidia_host_libraries.sh AND check in SitePackage.lua are broken


casparvl:
@bedroge and myself hit an issue on his CC 12.0 system where [my build for NCCL](https://github.com/EESSI/software-layer-scripts/pull/200#issuecomment-4243524549) failed with:

```
Warnings and errors of '/cvmfs/software.eessi.io/versions/2025.06/compat/linux/x86_64/usr/share/Lmod/libexec/lmod ...' command (stderr only):
Lmod has detected the following error:
Your driver CUDA version is 12.0
but the module you want to load requires CUDA 12.8.0. Please update your
CUDA driver libraries and then let EESSI know about the update.
For more information on how to do this, see
https://www.eessi.io/docs/site_specific_config/gpu/.

While processing the following module(s):
    Module fullname                             Module Filename
    ---------------                             ---------------
    UCX-CUDA/1.18.0-GCCcore-14.2.0-CUDA-12.8.0  /cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen5/accel/nvidia/cc120/modules/all/UCX-CUDA/1.18.0-GCCcore-14.2.0-CUDA-12.8.0.lua
```

We were surprised to see this, since the driver on the host is 13.0. Initially, we figured @bedroge just had to rerun the driver symlink script - but that didn't help. Even removing the `cuda_version.txt` from the `shared_fs` of the bot didn't help: it regenerated the `cuda_version.txt`, but with the same `12.0` content - rather than the expected `13.0`.

**BUG 1**

Turns out our driver version injected into `cuda_version.txt` is wrong. We query `nvidia-smi --query-gpu=gpu_name,count,driver_version,compute_cap` [here](https://github.com/EESSI/software-layer-scripts/blob/88653852578d0cc79a9fbc6b023cd968f0451b57/scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh#L224-L234), but `compute_cap` returns the compute capability of the hardware - not of the driver. Since @bedroge has RTX6000 cards, it returns `12.0`. I checked on my system, there the `cuda_version.txt` indeed contained `9.0` (for our H100 cards).

**BUG 2**

[...]

PR #200, opened 2026-04-14
Change 120f to 120 for cuda 12.8.0

casparvl:
Fixes

```
nvcc fatal   : Unsupported gpu architecture 'compute_120f'
```

Errors for the `120f` target and CUDA toolkit 12.8.0.

PR #202, opened 2026-04-14
Make cuda version check dynamic

### LINKED ISSUE #189: Get CUDA driver version dynamically
Currently, we store a small file to keep track of the CUDA driver version that was symlinked:

```
$ cat /cvmfs/software.eessi.io/versions/2025.06/compat/linux/x86_64/lib/nvidia/cuda_version.txt
9.0
```

The disadvantage is that if a new driver is installed, the host site needs to rerun `link_nvidia_host_libraries.sh`, even if the driver _location_ hasn't changed. I.e. the symlinks don't need any update - just this hardcoded driver version. To me, this feels like an unnecessary burdon on the host site to have to remember to rerun this every time.

If we could somehow get the driver version dynamically, the _only_ reason a host site would need to rerun the script is if the driver location on the host ever changed - and that's something I'd consider rare (much rarer than a driver update).

**Option 1**
This print the version:
```
strings /cvmfs/software.eessi.io/versions/2025.06/compat/linux/x86_64/lib/nvidia/libcuda.so | grep "CUDA version"
{"format":1,"CUDA version":"13.1","driver":[535,550,570,575,580,590],"device":[1,2,7,8,9,10,11,12,13,14]}
CUDA forward compatibility is disabled, but CUDA version [%d.%d] does not match RM version [%d.%d].
```

Advantages:
- Easy

Disadvantages

- it takes about a second to execute (that's quite slow if you're just trying to load a module).
- You'd need to create a regex, and it's quite sensitive to changes. Of course there is no guarantee that this pattern is consistent accross versions, or even that this string will _always_ be there.

**Option 2**
We can use a tiny (CPU-only) program
```
$ cat driver_version.c
#include <cuda.h>
#include <stdio.h>

int main() {
    int driverVersion;
    cuDriverGetVersion(&driverVersion);
    printf("CUDA Driver Version: %d\n", driverVersion);
    return 0;
}
```
compiled with
```
gcc -o driver_version driver_version.c -lcuda
```
to dump the driver version. The advantage is that it relies on the runtime linker to find CUDA, and thus will print the version of whatever driver is _actually_ found by the runtime linker in the compatibility layer. That's good, because _that_ is the version we are interested in. Running it is very fast (O(ms)).

One question is how/when we compile this. At compile time, it needs a CUDA module to be loaded (because it needs the CUDA header). At runtime, it doesn't. 

Advantages:

- Always get the driver version of the one _actually_ found by the runtime linker
- Relies on stability in the API - but is a very basic API function. Unlikely to ever change.

Disadvantages:

- Compiling on the fly is difficult, because it needs the CUDA module to be already loaded (and this check runs prior to that).
- It may theoretically create a chicken-and-egg situation because you need some degree of compatibility (call signature of cuDriverGetVersion in the header needs to match the one in the driver), to do what is essentially a compatiblity check

**Option 3**

Just run `nvidia-smi --version`:

```
$ time nvidia-smi --version
NVIDIA-SMI version  : 590.48.01
NVML version        : 590.48
DRIVER version      : 590.48.01
CUDA Version        : 13.1
```
It takes ~15 ms on my current system. Fast, but maybe you still want to cache the result.

Advantage:

- No questions about compatiblity, as this comes packaged with the driver (as far as I know)

Disadvantage:

- Slightly slower, so maybe we need to cache the result so it never gets run more than once.

casparvl:
Fixes https://github.com/EESSI/software-layer-scripts/issues/189

First step to fixing https://github.com/EESSI/software-layer-scripts/issues/201

Unblocks https://github.com/EESSI/software-layer-scripts/pull/200 , which blocks https://github.com/EESSI/software-layer/pull/1462 which blocks https://github.com/EESSI/software-layer/pull/1453 ...

This PR can be modified by modifying the `create_lmodsitepackage.py` with something like:

```
                if not cudaVersion or cudaVersion == "" then
                    -- Hardcode for local testing
                    -- local eessi_prefix = os.getenv("EESSI_PREFIX")
                    local eessi_prefix = pathJoin('/home', 'casparl', 'EESSI', 'software-layer-scripts')
                    local script = pathJoin(eessi_prefix, 'scripts', 'gpu_support', 'nvidia', 'get_cuda_driver_version.sh')
```

### DIFF
--- a/create_lmodsitepackage.py
+++ b/create_lmodsitepackage.py
@@ -190,9 +190,7 @@ local function eessi_cuda_and_libraries_enabled_load_hook(t)
             if cudaDriverExists then
                 local cudaVersion = os.getenv("EESSI_CUDA_DRIVER_VERSION")
                 if not cudaVersion or cudaVersion == "" then
-                    -- Hardcode for local testing
-                    -- local eessi_prefix = os.getenv("EESSI_PREFIX")
-                    local eessi_prefix = pathJoin('/home', 'casparl', 'EESSI', 'software-layer-scripts')
+                    local eessi_prefix = os.getenv("EESSI_PREFIX")
                     local script = pathJoin(eessi_prefix, 'scripts', 'gpu_support', 'nvidia', 'get_cuda_driver_version.sh')
                     -- Check return code first. We don't want source_sh to raise an LmodError, we just print
                     -- an LmodWarning stating we couldn't do a proper version compatibility check
--- a/create_lmodsitepackage.py
+++ b/create_lmodsitepackage.py
@@ -192,8 +192,13 @@ local function eessi_cuda_and_libraries_enabled_load_hook(t)
                 if not cudaVersion or cudaVersion == "" then
                     local eessi_prefix = os.getenv("EESSI_PREFIX")
                     local script = pathJoin(eessi_prefix, 'scripts', 'gpu_support', 'nvidia', 'get_cuda_driver_version.sh')
-                    -- Check return code first. We don't want source_sh to raise an LmodError, we just print
-                    -- an LmodWarning stating we couldn't do a proper version compatibility check
+                    -- We cannot immedately use source_sh, since lmod has no way of catching a potential error
+                    -- and we don't want this to raise an LmodError just because nvidia-smi doesn't exist or
+                    -- doesn't print the right output (happens on a node with nvidia-smi but no driver installed).
+                    -- The only way to catch this is to source the script first with os.execute and make sure it
+                    -- returns with a zero exit code. Unfortunately, this means we have to run nvidia-smi twice, which
+                    -- is a bit slow. Since the result is then cached in the EESSI_CUDA_DRIVER_VERSION environment
+                    -- variable, this is probably acceptable
                     local r1, r2, r3 = os.execute("bash -c 'source " .. script .. "'")
                     local exit_code = 0
                     if type(r1) == "number" then
--- a/create_lmodsitepackage.py
+++ b/create_lmodsitepackage.py
@@ -194,8 +194,16 @@ local function eessi_cuda_and_libraries_enabled_load_hook(t)
                     local script = pathJoin(eessi_prefix, 'scripts', 'gpu_support', 'nvidia', 'get_cuda_driver_version.sh')
                     -- Check return code first. We don't want source_sh to raise an LmodError, we just print
                     -- an LmodWarning stating we couldn't do a proper version compatibility check
-                    local rc = os.execute("bash -c 'source " .. script .. "'")
-                    if rc == 0 then
+                    local r1, r2, r3 = os.execute("bash -c 'source " .. script .. "'")
+                    local exit_code = 0
+                    if type(r1) == "number" then
+                        -- Lua 5.1 or earlier, this is our exit code
+                        exit_code = r1
+                    else
+                        -- Lua 5.2 or later, r3 is our exit code
+                        exit_code = r3
+                    end
+                    if exit_code == 0 then
                         source_sh("bash", script)
                     end
                 end
--- a/create_lmodsitepackage.py
+++ b/create_lmodsitepackage.py
@@ -192,28 +192,13 @@ local function eessi_cuda_and_libraries_enabled_load_hook(t)
                 if not cudaVersion or cudaVersion == "" then
                     local eessi_prefix = os.getenv("EESSI_PREFIX")
                     local script = pathJoin(eessi_prefix, 'scripts', 'gpu_support', 'nvidia', 'get_cuda_driver_version.sh')
-                    -- We cannot immedately use source_sh, since lmod has no way of catching a potential error
-                    -- and we don't want this to raise an LmodError just because nvidia-smi doesn't exist or
-                    -- doesn't print the right output (happens on a node with nvidia-smi but no driver installed).
-                    -- The only way to catch this is to source the script first with os.execute and make sure it
-                    -- returns with a zero exit code. Unfortunately, this means we have to run nvidia-smi twice, which
-                    -- is a bit slow. Since the result is then cached in the EESSI_CUDA_DRIVER_VERSION environment
-                    -- variable, this is probably acceptable
-                    local r1, r2, r3 = os.execute("bash -c 'source " .. script .. "'")
-                    local exit_code = 0
-                    if type(r1) == "number" then
-                        -- Lua 5.1 or earlier, this is our exit code
-                        exit_code = r1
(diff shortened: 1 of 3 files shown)


create_lmodsitepackage.py
```diff
@@ -189,28 +188,48 @@
...
-                if not cudaVersion then
-                    LmodError("No CUDA version file\\n" .. cudaVersionFile .. "\\nfound. " .. refer_to_docs)
+                local cudaVersion = os.getenv("EESSI_CUDA_DRIVER_VERSION")
+                if not cudaVersion or cudaVersion == "" then
+                    local eessi_prefix = os.getenv("EESSI_PREFIX")
+                    local script = pathJoin(eessi_prefix, 'scripts', 'gpu_support', 'nvidia', 'get_cuda_driver_version.sh')
+                    -- Check return code first. We don't want source_sh to raise an LmodError, we just print
+                    -- an LmodWarning stating we couldn't do a proper version compatibility check
+                    local rc = os.execute("bash -c 'source " .. script .. "'")
+                    if rc == 0 then
```
bedroge: For me this does not seem to work. Even though the script exists, the condition always evaluates to false. By adding some debugging statements, I found that `rc` is a boolean (or nil?), and `if rc then` works for me.

bedroge: https://www.lua.org/pil/22.2.html is not really clear, but my AI friend told me:

> In Lua, os.execute() returns different values depending on the Lua version.
> 
> ✅ Lua 5.2 and newer (5.2, 5.3, 5.4)
> 
> os.execute(command) returns three values:
> success, exit_type, code = os.execute("your_command")
> 
> Meaning:
> success → true if the command terminated normally, nil otherwise
> exit_type → string:
>   "exit" → program exited normally
>   "signal" → program was killed by a signal (Unix)
> code → numeric:
>   exit status (if "exit")
>   signal number (if "signal")

scripts/gpu_support/nvidia/get_cuda_driver_version.sh
bedroge, line 1: This file will have to be added to the following list to ensure that it gets deployed to cvmfs:
https://github.com/EESSI/software-layer-scripts/blob/main/install_scripts.sh#L210

create_lmodsitepackage.py
bedroge: Lua 5.1 apparently does only return a number...

> 💡 Portable pattern
> 
> If you want code that works across versions:
> ```
> local r1, r2, r3 = os.execute(cmd)
> 
> if type(r1) == "number" then
>     -- Lua 5.1
>     local exit_code = r1 / 256
> else
>     -- Lua 5.2+
>     local success = r1
>     local exit_code = r3
> end
> ```

casparvl: I'm confused, so I can catch 3 returns, even if the command only returns a single one (i.e. r1)? And this does not lead to errors?

Also: does this change than solve your original issue of this thing always evaluating to false? I guess I then just check `exit_code == 0` and that would work accross both, right?

bedroge: Yeah, apparently that works in Lua:
```
function test()
    return 42
end

> test()
42
> r1, r2, r3 = test()
> r1
42
> r2
nil
> r3
nil
```

Checking the exit code should work in that case, indeed.

scripts/gpu_support/nvidia/get_cuda_driver_version.sh
casparvl: Done

create_lmodsitepackage.py
casparvl: Done, can you check that the current version also works for you? (I tested it and it still works for me - I'm pretty sure I'm on Lua 5.1)

bedroge:
This now correctly sets `EESSI_CUDA_DRIVER_VERSION=13.1` on my RTX Pro 6000. 

Forcing it to e.g. `12.0` and loading something that needs CUDA 12.9.1 prints a nice error:
```
Your driver CUDA version is 12.5  but the module you want to load requires CUDA 12.9.1.
```
With `12.9` it works fine, as expected, as it doesn't check the patch version.

Just to test failures of the script, I modified the script and hardcoded an empty string for the driver version. Then I get the following, as expected:
```
Lmod Warning:  Environment variable EESSI_CUDA_DRIVER_VERSION not found. Cannot ensure that driver version is new enough for CUDA toolkit version: '12.9.1'. This module
will still be loaded, but may not function as expected. Export EESSI_CUDA_DRIVER_VERSION_SUPPRESS_WARNING=1 
```

Setting that env var allowed me to load the module.

create_lmodsitepackage.py
```diff
@@ -189,28 +188,46 @@
...
-                -- driver CUDA versions don't give a patch version for CUDA
-                local major, minor = string.match(cudaVersion, "(%d+)%.(%d+)")
-                local major_req, minor_req, patch_req = string.match(cudaVersion_req, "(%d+)%.(%d+)%.(%d+)")
-                local driver_libs_need_update = false
-                if major < major_req then
-                    driver_libs_need_update = true
-                elseif major == major_req then
-                    if minor < minor_req then
+                -- Account for the fact that the script sourced above was designed to never return a non-zero exit
+                -- even if it failes to set EESSI_CUDA_DRIVER_VERSION
```
bedroge:
```suggestion
                -- even if it fails to set EESSI_CUDA_DRIVER_VERSION
```

bedroge:

bedroge (approved):

bedroge:
Ingested!

scripts/gpu_support/nvidia/get_cuda_driver_version.sh
```diff
@@ -0,0 +1,6 @@
+# This can be leveraged by the source_sh() feature of Lmod
+# Because we want to source this without immediately raising an LmodError upon failure, this script
+# is designed to ALWAYS return a 0 exit code
+EESSI_CUDA_DRIVER_VERSION=$(nvidia-smi --query | grep -oP 'CUDA Version\s*:\s*\K[0-9.]+') || return 0
```
ocaisa, line 4: This doesn't allow for the fact that someone can use compatibility libraries to improve the compute capability supported by the driver (and this can be done in user space)

ocaisa:
```suggestion
EESSI_CUDA_DRIVER_VERSION=$(LD_LIBARARY_PATH=$EESSI_CUDA_COMPATIBILITY_LIBRARY_PATH nvidia-smi --query | grep -oP 'CUDA Version\s*:\s*\K[0-9.]+') || return 0
```
could give a way to do this

PR #204, opened 2026-04-16
CI check to avoid duplicate directory bot errors

agimenog:
Hi,

Same as https://github.com/EESSI/software-layer/pull/1464 but for software-layer-scripts

### DIFF
--- a/.github/workflows/bot_build.yml
+++ b/.github/workflows/bot_build.yml
@@ -9,20 +9,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout repo
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-
-      - name: Git clone software-layer
-        run: git clone https://github.com/EESSI/software-layer.git
+      - name: Git clone software-layer and software-layer-scripts inside it
+        run: |
+          git clone https://github.com/EESSI/software-layer.git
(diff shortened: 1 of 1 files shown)


.github/workflows/bot_build.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Git clone software-layer
+        run: git clone https://github.com/EESSI/software-layer.git
+
+      - name: Run symlink commands
+        run: |
+          set -e
+
+          for file in $(ls software-layer | egrep -v 'easystacks|LICENSE|README.md|^bot'); do
```
bedroge: I'm not sure if @casparvl had the same idea in https://gitlab.com/eessi/support/-/work_items/241, but I would basically run it in the exact same way as the one in software-layer:
- clone the software-layer repo
- clone software-layer-scripts in a subdir of software-layer
- run the symlinks commands as in software-layer's `bot/build.sh` script (i.e. with `ls software-layer-scripts` on this line)

This should ensure that the `ln` commands in software-layer always work. By having the same check on both sides, no duplicates can be introduced on either side.

agimenog: Good morning! Think already solved this, can you check? Thanks.

bedroge (approved):
Lgtm, thanks @agimenog!

PR #222, opened 2026-05-06
Allow different overrides for compat (i.e. CPU family) and modules (i.e. SOFTWARE_SUBDIR)

casparvl:
This faciliates e.g. checking of the available RISCV modules on an X86_64 host.

This is useful for e.g. running the test suite on an x86_64-based login node, when targetting RISCV-based batch nodes. The reason is that the test suite generates tests based on the available modules, so on the login node (where the reframe runtime runs), we want to see the RISCV modules. 

This change allows doing
```
EESSI_CPU_FAMILY_OVERRIDE=x86_64 EESSI_VERSION_OVERRIDE=2025.06-001 EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic module load EESSI/2025.06
```
on an X86_64-based host, and make the RISC-V module stack available:

```
$ module av

--------------------------------------------------------- /cvmfs/dev.eessi.io/riscv/versions/2025.06-001/software/linux/riscv64/generic/modules/all ---------------------------------------------------------
   Abseil/20250512.1-GCCcore-14.3.0              FFTW.MPI/3.3.10-gompi-2025a                         libidn2/2.3.8-GCCcore-14.3.0           (D)    Perl/5.38.0
...
```

Admittedly, the usage is a bit restricted since this is really only useful if module stacks are NOT in sync _and_ if the CPU family for the target of interest is different from the one on the host. Then again, it is an override: nothing changes in the default behavior, and you don't set it unless you have very good reasons to. It simply allows us extra flexibility when testing things.

Issue came up in https://github.com/EESSI/dev.eessi.io-riscv/issues/34#issuecomment-4387220708 when trying to set up period tests on the RISC-V cluster.

### DIFF
--- a/.github/workflows/tests_eessi_module.yml
+++ b/.github/workflows/tests_eessi_module.yml
@@ -338,17 +338,15 @@ jobs:
       - name: Test if we can override the CPU family to check risc-v modules on an x86 system
         run: |
           # Initialise Lmod
-          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/init/lmod/bash
-
-          # Unload EESSI
-          module purge
+          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/compat/linux/$(uname -m)/usr/share/Lmod/init/bash
+          
+          export MODULEPATH=init/modules
 
           # Set the overrides we want to test
           export EESSI_CPU_FAMILY_OVERRIDE=x86_64  # Running on an x86_64 system, so we need x86_64 compat layer
           export EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic  # Get modules for riscv64/generic
           export EESSI_VERSION_OVERRIDE=2025.06-001  # Override EESSI version, as versioning for riscv repo is different
-
-          # Load EESSI again with the overrides
+          
           module load EESSI/${{matrix.EESSI_VERSION}}
           
           # Make sure the compat layer works with a simple ls
--- a/.github/workflows/tests_eessi_module.yml
+++ b/.github/workflows/tests_eessi_module.yml
@@ -348,6 +348,12 @@ jobs:
           export EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic  # Get modules for riscv64/generic
           export EESSI_VERSION_OVERRIDE=2025.06-001  # Override EESSI version, as versioning for riscv repo is different
 
+          # Load EESSI again with the overrides
+          module load EESSI/${{matrix.EESSI_VERSION}}
+          
+          # Make sure the compat layer works with a simple ls
+          [ -n "${EPREFIX:-}" ] && "$EPREFIX/bin/ls" || { echo "EPREFIX is not defined" >&2; exit 1; }          
+          
           # Check that we have a riscv MODULEPATH
           if [[ "$MODULEPATH" == *"riscv"* ]]; then
               echo "MODULEPATH contains riscv prefix ($MODULEPATH)"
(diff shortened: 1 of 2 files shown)


ocaisa:
I was wondering how I had danced around this in my API PR: https://github.com/EESSI/api_data/pull/16/files#diff-9cf2000c53760d837a449f874e53f792819108d3a4bf346336d0f7d082deae2cR73-R78

ocaisa:
Looks good, can you just confirm that things still work fine _without_ the override set (just in case).

.github/workflows/tests_eessi_module.yml
```diff
@@ -316,6 +316,45 @@ jobs:
...
+          export EESSI_CPU_FAMILY_OVERRIDE=x86_64  # Running on an x86_64 system, so we need x86_64 compat layer
+          export EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic  # Get modules for riscv64/generic
+          export EESSI_VERSION_OVERRIDE=2025.06-001  # Override EESSI version, as versioning for riscv repo is different
+
+          # Check that we have a riscv MODULEPATH
+          if [[ "$MODULEPATH" == *"riscv"* ]]; then
+              echo "MODULEPATH contains riscv prefix ($MODULEPATH)"
+          else
+              exit 1
+          fi
```
ocaisa:
```suggestion
  check_riscv_cpu_override:
    runs-on: ubuntu-24.04
    strategy:
      fail-fast: false
      matrix:
        EESSI_VERSION:
          - '2025.06'

    steps:
      - name: Check out software-layer repository
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1

      - name: Mount EESSI CernVM-FS repository
        uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
        with:
          cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
          cvmfs_http_proxy: DIRECT
          cvmfs_repositories: software.eessi.io,dev.eessi.io

      - name: Test if we can override the CPU family to check risc-v modules on an x86 system
        run: |
          # Initialise Lmod
          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/init/lmod/bash

          # Unload EESSI
          module purge

          # Set the overrides we want to test
          export EESSI_CPU_FAMILY_OVERRIDE=x86_64  # Running on an x86_64 system, so we need x86_64 compat layer
          export EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic  # Get modules for riscv64/generic
          export EESSI_VERSION_OVERRIDE=2025.06-001  # Override EESSI version, as versioning for riscv repo is different

          # Check that we have a riscv MODULEPATH
          if [[ "$MODULEPATH" == *"riscv"* ]]; then
              echo "MODULEPATH contains riscv prefix ($MODULEPATH)"
          else
              exit 1
          fi
```

.github/workflows/tests_eessi_module.yml
```diff
@@ -316,6 +316,45 @@ jobs:
...
+
+          # Unload EESSI
+          module purge
+
+          # Set the overrides we want to test
+          export EESSI_CPU_FAMILY_OVERRIDE=x86_64  # Running on an x86_64 system, so we need x86_64 compat layer
+          export EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic  # Get modules for riscv64/generic
+          export EESSI_VERSION_OVERRIDE=2025.06-001  # Override EESSI version, as versioning for riscv repo is different
+
+          # Check that we have a riscv MODULEPATH
```
ocaisa:
```suggestion

          # Load EESSI again with the overrides
          module load EESSI/${{matrix.EESSI_VERSION}}
          
          # Make sure the compat layer works with a simple ls
          [ -n "${EPREFIX:-}" ] && "$EPREFIX/bin/ls" || { echo "EPREFIX is not defined" >&2; exit 1; }          
          
          # Check that we have a riscv MODULEPATH
```

ocaisa (approved):
LGTM

PR #223, opened 2026-05-06
Ignore ROCm-LLVM on aarch64

zerefwayne:
ROCm-LLVM 6.4.1 is not supported on aarch64 family of CPUs.

See: https://github.com/EESSI/software-layer/pull/1473#issuecomment-4370846033

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -846,10 +846,11 @@ def is_unsupported_module(self):
     if not os.getenv("EESSI_OVERRIDE_ROCM_VERSION_CHECK"):
         if self.cfg.name == 'ROCm-LLVM' and self.cfg.version in ('6.4.1',):
             if get_eessi_envvar('EESSI_CPU_FAMILY') == 'aarch64':
-                msg = "ROCm-LLVM/6.4.1 is not supported on aarch64 architectures. "
                 msg += "Building with '--module-only --force' and injecting an LmodError into the modulefile."
                 msg += "You can override this behaviour by setting the EESSI_OVERRIDE_ROCM_VERSION_CHECK environment variable."
                 print_warning(msg)
+                errmsg = "ROCm-LLVM/6.4.1 is not supported on the aarch64 architecture."
+                errmsg += "For more details, see: https://github.com/EESSI/software-layer/pull/1473#issuecomment-4370846033"
                 var=EESSI_IGNORE_AARCH64_ROCMLLVM641_ENVVAR
                 setattr(self, EESSI_UNSUPPORTED_MODULE_ATTR, UnsupportedModule(envvar=var, errmsg=errmsg))
                 return True
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -844,7 +844,7 @@ def is_unsupported_module(self):
     # ROCm-LLVM 6.4.1 is not supported on aarch64
     # see: https://github.com/EESSI/software-layer/pull/1473#issuecomment-4370846033
     if not os.getenv("EESSI_OVERRIDE_ROCM_VERSION_CHECK"):
-        if ec.name == 'ROCm-LLVM' and ec.version in ('6.4.1',):
+        if self.cfg.name == 'ROCm-LLVM' and self.cfg.version in ('6.4.1',):
             if get_eessi_envvar('EESSI_CPU_FAMILY') == 'aarch64':
                 msg = "ROCm-LLVM/6.4.1 is not supported on aarch64 architectures. "
                 msg += "Building with '--module-only --force' and injecting an LmodError into the modulefile."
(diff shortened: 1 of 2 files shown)


ocaisa (commented):
Why would we do this? If it is not supported at all for Arm then why create modules?

The EESSI module will not expose a `MODULEPATH` that does not exist. What I would suggest is that we instead print a message in the EESSI module when we see aarch64 and an AMD GPU.

ocaisa:
What I mean is, this shouldn't be a warning but an error.

ocaisa:
My assumption here is `ROCm-LLVM` is being installed under an accelerator path. If that is not the case then I think we should discuss that choice.

eb_hooks.py
```diff
@@ -840,6 +841,19 @@ def is_unsupported_module(self):
...
+    # see: https://github.com/EESSI/software-layer/pull/1473#issuecomment-4370846033
+    if not os.getenv("EESSI_OVERRIDE_ROCM_VERSION_CHECK"):
+        if self.cfg.name == 'ROCm-LLVM' and self.cfg.version in ('6.4.1',):
+            if get_eessi_envvar('EESSI_CPU_FAMILY') == 'aarch64':
+                msg = "ROCm-LLVM/6.4.1 is not supported on aarch64 architectures. "
+                msg += "Building with '--module-only --force' and injecting an LmodError into the modulefile."
+                msg += "You can override this behaviour by setting the EESSI_OVERRIDE_ROCM_VERSION_CHECK environment variable."
+                print_warning(msg)
+                var=EESSI_IGNORE_AARCH64_ROCMLLVM641_ENVVAR
+                setattr(self, EESSI_UNSUPPORTED_MODULE_ATTR, UnsupportedModule(envvar=var, errmsg=errmsg))
```
casparvl: We should define a sensible `errmsg` before this line. Note that this is the `errmsg` that gets printed to the _end user_ trying to load the module. Thus, it should make sense to such an end-user...

casparvl:
> My assumption here is ROCm-LLVM is being installed under an accelerator path. If that is not the case then I think we should discuss that choice.

It is.

> What I mean is, this shouldn't be a warning but an error.

You get a warning at install time that this will trigger a `--module-only` install. The module that this will install _will_ raise an `LmodError` when you try to load it. The `is_unsupported_module` function in `eb_hooks.py` provides standard functionality to put such modules in place. I.e. it triggers a `--module-only` installation, and appends an LmodError to the module with a message that is configurable. This makes it very easy to add other cases of 'unsupported modules'. Note that the original case for which we implemented something like this (though not in this generic form yet) was the fact that zen4 didn't support the GCCcore 12.2-based toolchains.

> Why would we do this? If it is not supported at all for Arm then why create modules?

> The EESSI module will not expose a MODULEPATH that does not exist. What I would suggest is that we instead print a message in the EESSI module when we see aarch64 and an AMD GPU.

This is another option. I'm ok with this as well. It depends a bit on your philosophy: if we want all module environments to look identical in terms of which modules are present, we should take the `is_unsupported_module` approach. The downside is that I'm not sure what we look at for our software overview / API (and maybe you can tell me more about this :)): if we look at the modules, then it would _appear_ that these installations are present for AArch64+AMD GPU combinations, while actually they are not.

One other reason I'd have for opting for the `is_unsupported_module` approach is that it is more consistent: we did the same thing for GCCcore-12.2.0-based stuff not being supported on `zen4`. We _could_ have added a generic warning in the EESSI module for that as well, and simply never built those modules for the zen4-target - but we decided to put modules in place that print errors. I'm not sure why the current situation would be _different_?

ocaisa:
Think of it in terms of what usable modules are in the `MODULEPATH`. For the Zen4 case, it's a CPU and you have all the other toolchains that work just fine, so it made sense to fill the holes. For aarch64+ROCm there would be no usable modules at all, so there really isn't any point in going to the effort of creating that `MODULEPATH` (and the EESSI module handles that absence just fine)

init/modules/EESSI/2023.06.lua
```diff
@@ -180,6 +180,9 @@ if not (archdetect_accel == nil or archdetect_accel == '') then
     -- /cvmfs/software.eessi.io/versions/<EESSI_VERSION>/software/linux/x86_64/amd/zen3/accel/nvidia/cc80/modules/all
     eessi_module_path_accel = pathJoin(eessi_accel_software_path, archdetect_accel, eessi_modules_subdir)
     eessiDebug("Checking if " .. eessi_module_path_accel .. " exists")
+    if eessi_cpu_family == "aarch64" and archdetect_accel:match("^accel/amd/") then
+        LmodMessage("aarch64 CPU detected, AMD ROCm doesn't support aarch64 yet")
```
casparvl:
```suggestion
        LmodMessage("AArch64 CPU architecture and AMD GPU detected. However, AMD ROCm does not support AArch64. Thus, AMD GPU-enabled EESSI installations are not available on this system.")
```

PR #227, opened 2026-05-08
Fix EESSI-extend CI by doing a `module --force purge`

bedroge:
Seems required because https://github.com/EESSI/github-action-eessi/pull/36 made the EESSI module sticky.


ocaisa (approved):
Thanks for fixing this

PR #228, opened 2026-05-12
Add gpu tag for ROCm-LLVM and refactor method

zerefwayne:
Easyconfigs which include ROCm-LLVM as a dependency should also be tagged with `gpu`. It doesn't need to be dropped to a build dependency as it is redistributable (unlike CUDA).

The loop seems to iterate over dependencies twice, it can be simplified to one pass.

### DIFF
--- a/.github/workflows/test-eb-hooks.yml
+++ b/.github/workflows/test-eb-hooks.yml
@@ -116,3 +116,55 @@ jobs:
               eb --hooks=$PWD/eb_hooks.py "$INCOMPATIBLE_EASYCONFIG" --stop fetch 2>&1 1>/dev/null | grep -q "does not contain a valid list of dictionaries"
               echo "Incorrect format for EESSI_SITE_TOP_LEVEL_TOOLCHAINS caught"
 
+  check_inject_gpu_property:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        # ROCm-LLVM / rocm-compilers / rompi only exist in the 2025.06 stack
+        EESSI_VERSION:
+          - '2025.06'
+
+    steps:
+        - name: Check out software-layer repository
+          uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+        - name: Mount EESSI CernVM-FS repository
+          uses: eessi/github-action-eessi@v3
+          with:
+            eessi_stack_version: ${{matrix.EESSI_VERSION}}
+            use_eessi_module: true
+
+        - name: Test that inject_gpu_property tags GPU software correctly
+          if: ${{ github.event_name == 'pull_request' }}
+          run: |
+              module load EESSI-extend/${{matrix.EESSI_VERSION}}-easybuild
+
+              echo "Running inject_gpu_property tests for EESSI/${{matrix.EESSI_VERSION}}..."
+
+              # Control: a non-GPU easyconfig must not receive the gpu property
+              eb --hooks=$PWD/eb_hooks.py --extended-dry-run M4-1.4.19-GCCcore-14.2.0.eb 2>&1 | grep -qiF 'add_property("arch","gpu")' && exit 1
+              echo "Non-GPU easyconfig untouched"
+
+              # ROCm-LLVM as a direct dependency (rocm-compilers bundle)
+              echo "Testing ROCm-LLVM as a direct dependency"
+              OUT=$(eb --hooks=$PWD/eb_hooks.py --extended-dry-run rocm-compilers-19.0.0-ROCm-6.4.1.eb 2>&1)
+              echo "$OUT" | grep -qiF 'add_property("arch","gpu")'
+              echo "$OUT" | grep -qiF 'setenv("EESSIROCMVERSION","6.4.1")'
+              echo "Direct ROCm-LLVM dependency detected"
+
+              # rocm-compilers as the toolchain
+              echo "Testing ROCm-LLVM in a rocm-compilers toolchain"
+              OUT=$(eb --hooks=$PWD/eb_hooks.py --extended-dry-run rocBLAS-4.4.0-rocm-compilers-19.0.0-ROCm-6.4.1.eb 2>&1)
+              echo "$OUT" | grep -qiF 'add_property("arch","gpu")'
+              echo "$OUT" | grep -qiF 'setenv("EESSIROCMVERSION","6.4.1")'
+              echo "rocm-compilers toolchain handled"
+
+              # TODO: rompi test to be brought back once rompi/2025a is available in EESSI
+
+              # # rompi as the toolchain (ROCm-LLVM nested inside rocm-compilers bundle)
+              # echo "Testing ROCm-LLVM in a rompi toolchain (rocm-compilers bundle)"
+              # # added --disable-cleanup-tmpdir to ignore irrelevant EasyBuild cleanup error
+              # OUT=$(eb --hooks=$PWD/eb_hooks.py --extended-dry-run OSU-Micro-Benchmarks-7.5-rompi-2025a.eb 2>&1)
+              # echo "$OUT" | grep -qiF 'add_property("arch","gpu")'
+              # echo "$OUT" | grep -qiF 'setenv("EESSIROCMVERSION","6.4.1")'
+              # echo "rompi toolchain handled"
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -1944,34 +1944,43 @@ def inject_gpu_property(ec):
     # - drop dependency to build dependency
     # - add 'gpu' Lmod property
     # - add envvar with package version
-    pkg_names = ( "CUDA", "cuDNN" )
+    pkg_names = ( "CUDA", "cuDNN", "ROCm-LLVM" )
+    # skip dropping step for ROCm-LLVM as it is redistributable
+    drop_to_builddep = ( "CUDA", "cuDNN" )
     pkg_versions = { }
     add_gpu_property = ''
 
+    # If none of the gpu packages are in the easyconfig, do not process further
+    dep_names = {dep[0] for dep in ec_dict['dependencies']}
+    if not any(pkg in dep_names for pkg in pkg_names):
+        return ec
+
+    # Check if pkg_name is in the dependencies, if so drop dependency to build
+    # dependency and set variable for later adding the 'gpu' Lmod property
+    # to '.remove' dependencies from ec_dict['dependencies'] we make a copy,
+    # iterate over the copy and can then savely use '.remove' on the original
     for pkg_name in pkg_names:
-        # Check if pkg_name is in the dependencies, if so drop dependency to build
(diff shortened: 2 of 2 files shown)


eb_hooks.py
```diff
@@ -1944,34 +1944,43 @@ def inject_gpu_property(ec):
...
     # - add envvar with package version
-    pkg_names = ( "CUDA", "cuDNN" )
+    pkg_names = ( "CUDA", "cuDNN", "ROCm-LLVM" )
+    # skip dropping step for ROCm-LLVM as it is redistributable
+    drop_to_builddep = ( "CUDA", "cuDNN" )
     pkg_versions = { }
     add_gpu_property = ''
 
+    # If none of the gpu packages are in the easyconfig, do not process further
+    dep_names = {dep[0] for dep in ec_dict['dependencies']}
```
ocaisa: Have you tested this? I wonder if ROCm-LLVM is the toolchain if it actually appears in the dependencies?

casparvl: I doubt it. For one of the other hooks (can't remember which) where we searched some dep list with CUDA / cuDNN in it, I added logic to also check the toolchain instead - and then just trigger it if it's `rocm-compilers`, `rompi`,... etc. This should probably do something similar. @zerefwayne search for `rocm-compilers` in `eb_hooks.py` and you'll probably find my logic in the other hook.

ocaisa: It would be better if we can identify ROCm-LLVM explicitly so that we can store the version in an environment variable. That will be useful later in Lmod when we are checking if the version is supported by the driver.

ocaisa: I don't know if you have a toolchain instance at this point but if you did you could use it to find the ROCm-LLVM versionsuffix
```
ec.toolchain.dependencies()
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
+            if dep not in ec_dict['builddependencies']:
+                ec_dict['builddependencies'].append(dep)
+
+    # ROCm-LLVM is handled separately: it is redistributable (kept as a runtime dep)
+    # and may be pulled in via a ROCm toolchain rather than as a direct dependency.
+    rocm_llvm_dep = find_rocm_llvm_dependency(ec)
+    if rocm_llvm_dep is not None:
+        add_gpu_property = 'add_property("arch","gpu")'
+        versionsuffix = rocm_llvm_dep[2] if len(rocm_llvm_dep) > 2 else ''
+        pkg_versions['ROCm-LLVM'] = rocm_llvm_dep[1] + versionsuffix
```
ocaisa: For this, I think it is not more interesting to store the ROCm version as later we need something that we can compare with output we can get from `rocm_smi` (like we do for `nvidia_smi` and the CUDA version)
```suggestion
        rocm_prefix = "-ROCm-"

        if versionsuffix.startswith(rocm_prefix):
            rocm_version = versionsuffix[len(rocm_prefix):]
        else:
            raise EasyBuildError(f"Invalid format for ROCm versionssuffix: {versionsuffix}")
        pkg_versions['ROCm'] = rocm_version
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
 
 
+def find_rocm_llvm_dependency(ec):
+    """
+    Return the ROCm-LLVM dependency for this easyconfig, or None. ROCm-LLVM can
+    be a direct dependency, a direct toolchain component (rocm-compilers as the
+    toolchain), or one level deeper inside the rocm-compilers bundle when the
+    toolchain is rompi/rfbf/rfoss.
+    """
+    for dep in ec.asdict()['dependencies']:
```
casparvl, line 2028:
```suggestion
    # Check if ROCm-LLVM is a direct dependency, and if so, return that
    for dep in ec.asdict()['dependencies']:
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
+        if dep[0] == 'ROCm-LLVM':
+            return dep
+
+    # Only inspect the toolchain for ROCm toolchains.
```
casparvl:
```suggestion
    # ROCm-LLVM can also be part of the toolchain. First, return early if this is NOT a ROCm-based toolchain
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
+    if ec['toolchain']['name'] not in ('rocm-compilers', 'rompi', 'rfbf', 'rfoss'):
+        return None
+
+    tcdeps = ec.toolchain.tcdeps or []
+    for dep in tcdeps:
```
casparvl, line 2045:
```suggestion
    # Check if ROCm-LLVM is a direct dependency for this toolchain (which would be the case for rocm-compilers)
    for dep in tcdeps:
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
+        if dep[0] == 'ROCm-LLVM':
+            return dep
+    for dep in tcdeps:
```
casparvl, line 2051:
```suggestion
    # For rompi, rfbf, rfoss, ROCm-LLVM is an indirect dependency, via rocm-compilers
    # First, confirm that this toolchain has rocm-compilers as direct dependency
    for dep in tcdeps:
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
+        if dep[0] == 'rocm-compilers':
+            versionsuffix = dep[2] if len(dep) > 2 else ''
```
casparvl:
```suggestion
            # Now, we need information from rocm-compilers, but that is NOT present in the current 'ec' instance
            # We need to find the easyconfig for rocm-compilers with robot_find_easyconfig, then parse it using process_easyconfig
            # Finally, we can get the (runtime) dependencies from that easyconfig and return the ROCm-LLVM dependency
            versionsuffix = dep[2] if len(dep) > 2 else ''
```

eb_hooks.py
casparvl: Agreed, this makes sense to me

casparvl (changes requested):
I think this PR is in decent shape. 

The `find_rocm_llvm_dependency` function could use a few more comments to make it easier to understand what's going on. I've made some suggestion. 

In addition, I agree with Alan's suggestion to store the ROCm version, not the full ROCm-LLVM version+versionsuffix. That'll save us some parsing later on, and it's unlikely we'll be interested in the LLVM version anyway. That will also require an update of the `test-eb-hooks.py`, see my suggestions.

If you can apply those changes, I think this is good to go. Note that this is blocking progress on https://github.com/EESSI/software-layer-scripts/pull/231 , which in turn is blocking progress on our `2026.06` compatibility layer. So I'd like to get this one merged ASAP :)

.github/workflows/test-eb-hooks.yml
```diff
@@ -116,3 +116,48 @@ jobs:
...
+              module load EESSI-extend/${{matrix.EESSI_VERSION}}-easybuild
+
+              # Control: a non-GPU easyconfig must not receive the gpu property
+              eb --hooks=$PWD/eb_hooks.py --extended-dry-run M4-1.4.19-GCCcore-14.2.0.eb 2>&1 | grep -qF 'add_property("arch","gpu")' && exit 1
+              echo "Non-GPU easyconfig untouched"
+
+              # ROCm-LLVM as a direct dependency (rocm-compilers bundle)
+              OUT=$(eb --hooks=$PWD/eb_hooks.py --extended-dry-run rocm-compilers-19.0.0-ROCm-6.4.1.eb 2>&1)
+              echo "$OUT" | grep -qF 'add_property("arch","gpu")'
+              echo "$OUT" | grep -qF 'setenv("EESSIROCMMINLLVMVERSION","19.0.0-ROCm-6.4.1")'
```
casparvl:
```suggestion
              echo "$OUT" | grep -qF 'setenv("EESSIROCMVERSION","6.4.1")'
```

.github/workflows/test-eb-hooks.yml
```diff
@@ -116,3 +116,48 @@ jobs:
...
+              echo "Direct ROCm-LLVM dependency detected"
+
+              # rocm-compilers as the toolchain
+              OUT=$(eb --hooks=$PWD/eb_hooks.py --extended-dry-run hwloc-ROCm-2.11.2-rocm-compilers-19.0.0-ROCm-6.4.1.eb 2>&1)
+              echo "$OUT" | grep -qF 'add_property("arch","gpu")'
+              echo "$OUT" | grep -qF 'setenv("EESSIROCMMINLLVMVERSION","19.0.0-ROCm-6.4.1")'
```
casparvl:
```suggestion
              echo "$OUT" | grep -qF 'setenv("EESSIROCMVERSION","6.4.1")'
```

.github/workflows/test-eb-hooks.yml
```diff
@@ -116,3 +116,48 @@ jobs:
...
+              echo "rocm-compilers toolchain handled"
+
+              # rompi as the toolchain (ROCm-LLVM nested inside rocm-compilers bundle)
+              OUT=$(eb --hooks=$PWD/eb_hooks.py --extended-dry-run OSU-Micro-Benchmarks-7.5-rompi-2025a.eb 2>&1)
+              echo "$OUT" | grep -qF 'add_property("arch","gpu")'
+              echo "$OUT" | grep -qF 'setenv("EESSIROCMMINLLVMVERSION","19.0.0-ROCm-6.4.1")'
```
casparvl:
```suggestion
              echo "$OUT" | grep -qF 'setenv("EESSIROCMVERSION","6.4.1")'
```

eb_hooks.py
```diff
@@ -1934,44 +1938,80 @@ def replace_binary_non_distributable_files_with_symlinks(log, install_dir, pkg_n
...
+        pkg_versions['ROCm-LLVM'] = rocm_llvm_dep[1] + versionsuffix
+
     if add_gpu_property:
         ec.log.info("Injecting gpu as Lmod arch property and envvars for dependencies with their version")
         modluafooter = 'modluafooter'
         extra_mod_footer_lines = [add_gpu_property]
         for pkg_name, version in pkg_versions.items():
-            envvar = "EESSI%sVERSION" % pkg_name.upper()
+            # follow the convention for replacing - with MIN
+            envvar = "EESSI%sVERSION" % pkg_name.upper().replace('-', 'MIN')
```
casparvl: No longer needed if we don't use `ROCm-LLVM` but `ROCm` as package name.

```suggestion
            envvar = "EESSI%sVERSION" % pkg_name.upper()
```

casparvl (approved):

PR #230, opened 2026-05-13
Add toolchain specific RPATH in host_injections

TopRichard:
This may be relevant for EESSI/2026.x or when building software locally on top of EESSI. It introduces an additional path for MPI injection per toolchain version, while preserving the existing generic path.

Further testing is required.

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -233,7 +233,7 @@ def get_eessi_envvar(eessi_envvar):
     return eessi_envvar_value
 
 
-def get_rpath_override_dirs(software_name):
+def get_rpath_override_dirs(software_name, toolchain_version):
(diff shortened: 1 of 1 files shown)


eb_hooks.py
```diff
@@ -233,7 +233,7 @@ def get_eessi_envvar(eessi_envvar):
     return eessi_envvar_value
 
 
-def get_rpath_override_dirs(software_name):
+def get_rpath_override_dirs(software_name, toolchain_version):
```
ocaisa, line 236: Not sure this is what you want, the toolchain version is not specific enough since you also need to cover  CUDA versions.

TopRichard: True, while this does not directly address https://github.com/EESSI/software-layer-scripts/issues/220, it does make it possible to use a different CUDA version for each toolchain version.

PR #232, opened 2026-05-19
Update site package behavior

casparvl:
This is preparatory work for better supporting building on top of EESSI. 

Specifically, this change allows sites to customize where their site installation path is through `EESSI_SITE_SOFTWARE_PREFIX`. It also makes sure that any caches for the site-installs are picked up by adding the `lmodrc` file to the `LDOD_RC` search path. Thus, as long as sites generate an `lmodrc.lua` (e.g. using the `create_lmodrc.py` from the `software-layer-scripts` repo) _and_ run the appropriate commands to generate a cache in the corresponding `.lmod` folder (analogous to how we do this for EESSI), this will allow them to also have a proper cache for locally installed modules.

An example would be:

```
[casparl@int5 software-layer-scripts]$ export EESSI_SITE_SOFTWARE_PREFIX=/foo/bar
[casparl@int5 software-layer-scripts]$ module load EESSI/2025.06
Module for EESSI/2025.06 loaded successfully
[casparl@int5 software-layer-scripts]$ echo $EESSI_SITE_SOFTWARE_PATH
/foo/bar/versions/2025.06/software/linux/x86_64/amd/zen2
[casparl@int5 software-layer-scripts]$ echo $MODULEPATH
/foo/bar/versions/2025.06/software/linux/x86_64/amd/zen2/modules/all:/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/modules/all:/gpfs/home4/casparl/EESSI/software-layer-scripts/init/modules
[casparl@int5 software-layer-scripts]$ echo $LMOD_RC
/foo/bar/versions/2025.06/software/linux/x86_64/amd/zen2/.lmod/lmodrc.lua:/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/.lmod/lmodrc.lua
```

### DIFF
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -119,7 +119,20 @@ if [ -d $EESSI_PREFIX ]; then
         if [ ! -z $EESSI_BASIC_ENV ]; then
           show_msg "Only setting up basic environment, so we're done"
         elif [ -d $EESSI_SOFTWARE_PATH ]; then
-          export EESSI_SITE_SOFTWARE_PATH=${EESSI_SOFTWARE_PATH/versions/host_injections}
+          if [ -z ${EESSI_SITE_SOFTWARE_PREFIX} ]; then
+            site_software_path=${EESSI_SOFTWARE_PATH/versions/host_injections}
+          else
+            site_software_path=${EESSI_SOFTWARE_PATH/$EESSI_SITE_SOFTWARE_PREFIX/$EESSI_CVMFS_REPO}
+          fi
+          export EESSI_SITE_SOFTWARE_PATH=${site_software_path}
+          # Prepend site's lmodrc.lua path to $LMOD_RC
+          show_msg "Adding Lmod site configuration file path ($site_lmod_rc_file) to $LMOD_RC"
+          site_lmod_rc_file="$EESSI_SITE_SOFTWARE_PATH/.lmod/lmodrc.lua"
+          if [ -z ${LMOD_RC} ]; then
+            export LMOD_RC=$site_lmod_rc_file
+          else
+            export LMOD_RC=$site_lmod_rc_file:$LMOD_RC
+          fi
           show_msg "Using ${EESSI_SITE_SOFTWARE_PATH} as the site extension directory for installations."
           EESSI_SITE_ACCEL_SOFTWARE_PATH=${EESSI_ACCEL_SOFTWARE_PATH/versions/host_injections}
           show_msg "Using ${EESSI_SITE_ACCEL_SOFTWARE_PATH} as the site extension directory for accelerated installations."
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -121,11 +121,16 @@ if [ -d "$EESSI_PREFIX" ]; then
         elif [ -d "$EESSI_SOFTWARE_PATH" ]; then
           if [ -z "${EESSI_SITE_SOFTWARE_PREFIX}" ]; then
             site_software_path="${EESSI_SOFTWARE_PATH/versions/host_injections}"
+            site_software_accel_path="${EESSI_ACCEL_SOFTWARE_PATH/versions/host_injections}"
           else
             site_software_path="${EESSI_SOFTWARE_PATH/$EESSI_SITE_SOFTWARE_PREFIX/$EESSI_CVMFS_REPO}"
+            site_software_accel_path="${EESSI_ACCEL_SOFTWARE_PATH/$EESSI_SITE_SOFTWARE_PREFIX/$EESSI_CVMFS_REPO}"
           fi
           export EESSI_SITE_SOFTWARE_PATH="${site_software_path}"
-          show_msg "Set the site software path to $EESSI_SITE_SOFTWARE_PATH"
+          EESSI_SITE_ACCEL_SOFTWARE_PATH="${site_software_accel_path}"
+          show_msg "Using ${EESSI_SITE_SOFTWARE_PATH} as the site extension directory for installations."
+          show_msg "Using ${EESSI_SITE_ACCEL_SOFTWARE_PATH} as the site extension directory for accelerated installations."
+
           # Prepend site's lmodrc.lua path to $LMOD_RC
           show_msg "Adding Lmod site configuration file path ($site_lmod_rc_file) to $LMOD_RC"
           site_lmod_rc_file="$EESSI_SITE_SOFTWARE_PATH/.lmod/lmodrc.lua"
@@ -134,9 +139,6 @@ if [ -d "$EESSI_PREFIX" ]; then
           else
             export LMOD_RC="$site_lmod_rc_file:$LMOD_RC"
           fi
-          show_msg "Using ${EESSI_SITE_SOFTWARE_PATH} as the site extension directory for installations."
-          EESSI_SITE_ACCEL_SOFTWARE_PATH="${EESSI_ACCEL_SOFTWARE_PATH/versions/host_injections}"
-          show_msg "Using ${EESSI_SITE_ACCEL_SOFTWARE_PATH} as the site extension directory for accelerated installations."
           # Allow for use of alternative module tree shipped with EESSI
           if [ -z ${EESSI_MODULE_SUBDIR+x} ]; then
             # EESSI_MODULE_SUBDIR not set
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -18,11 +18,11 @@ function show_msg {
 # set up minimal environment: $EESSI_PREFIX, $EESSI_VERSION, $EESSI_OS_TYPE, $EESSI_CPU_FAMILY, $EPREFIX
 source $EESSI_INIT_DIR_PATH/minimal_eessi_env
 
-if [ -d $EESSI_PREFIX ]; then
+if [ -d "$EESSI_PREFIX" ]; then
   show_msg "Found EESSI repo @ $EESSI_PREFIX!"
 
-  export EESSI_EPREFIX=$EPREFIX
-  if [ -d $EESSI_EPREFIX ]; then
+  export EESSI_EPREFIX="$EPREFIX"
+  if [ -d "$EESSI_EPREFIX" ]; then
 
     # determine subdirectory in software layer
     if [ "$EESSI_USE_ARCHDETECT" == "1" ]; then
@@ -32,8 +32,8 @@ if [ -d $EESSI_PREFIX ]; then
       #   under $EESSI_PREFIX/software/$EESSI_OS_TYPE; if so use the architecture as best match
       IFS=: read -r -a archs <<< "${all_cpupaths}"
       for arch in "${archs[@]}"; do
-        if [ -d ${EESSI_PREFIX}/software/${EESSI_OS_TYPE}/${arch} ]; then
-          export EESSI_SOFTWARE_SUBDIR=${arch}
+        if [ -d "${EESSI_PREFIX}/software/${EESSI_OS_TYPE}/${arch}" ]; then
+          export EESSI_SOFTWARE_SUBDIR="${arch}"
           show_msg "archdetect says ${EESSI_SOFTWARE_SUBDIR}"
           break
         fi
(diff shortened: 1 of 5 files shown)


init/modules/EESSI/2023.06.lua
```diff
@@ -134,7 +134,16 @@ local eessi_modules_subdir = pathJoin("modules", "all")
 -- eessi_module_path is the location of the _CPU_ module files, e.g.,
 -- /cvmfs/software.eessi.io/versions/<EESSI_VERSION>/software/linux/x86_64/amd/zen3/modules/all
 local eessi_module_path = pathJoin(eessi_software_path, eessi_modules_subdir)
-local eessi_site_software_path = string.gsub(eessi_software_path, "versions", "host_injections")
+local eessi_site_software_path
+-- If EESSI_SITE_SOFTWARE_PREFIX is defined, replace /cvmfs/software.eessi.io (or more generally EESSI_CVMFS_REPO)
+-- by that prefix. This ensures that the directory still contains the os/vendor/arch/micro-arch/accelerator etc
+-- If it is not defined, default to a site installation prefix under host_injections
+site_prefix = os.getenv("EESSI_SITE_SOFTWARE_PREFIX")
```
ocaisa, line 141: I don't love this approach, there is value in having the path in a fully defined location.

ocaisa: I thought about it more, and I am coming around. I was a bit stuck on having the fixed path because we require a fixed path to be able to do MPI injection. However, these are actually separate issues and this change doesn't affect that.

ocaisa: This will require some documentation updates, and some additional checks in `EESSI-extend`. We should also update the dev.eessi.io workflow to use this. 

Issues for all these are enough for now, but they should be addressed.

init/modules/EESSI/2023.06.lua
```diff
@@ -169,8 +178,14 @@ if ( mode() ~= "spider" ) then
     prepend_path("MODULEPATH", eessi_module_path)
     eessiDebug("Adding " .. eessi_module_path .. " to MODULEPATH")
 end
+
+-- Make sure the EESSI cache is found, this is specified in the lmodrc.lua in the eessi_software_path
 prepend_path("LMOD_RC", pathJoin(eessi_software_path, ".lmod", "lmodrc.lua"))
 eessiDebug("Adding " .. pathJoin(eessi_software_path, ".lmod", "lmodrc.lua") .. " to LMOD_RC")
+-- Make sure that a cache for site installations can also be found
```
ocaisa, line 185: How to generate the cache also needs an update, but I need to check that works first. We need to let Lmod know that the module path requires a gateway module, then the hierarchy can be represented in the cache.

It's a separate issue, this just reminds me that I need to look into it.

init/modules/EESSI/2023.06.lua
bedroge: Caspar and I discussed this quite a bit yesterday. In some way I also still prefer to have everything in a single and fixed location, but I also see Caspar's point, and being able to do the software installations elsewhere does make sense to me. I still have some concerns, e.g. about where to put the Lmod site hooks: Caspar wanted to keep them under host_injections, as they (may) also affect EESSI modules, and though that makes sense, it does mean that you may have two locations with this `2023.06/software/x86_64/amd/zen3` trees (one for locally built software, one for Lmod hooks). Same question for the site's Lmod cache: for that one it does definitely make sense to store them near the software, meaning you will end up with two of those `.lmod` directories in different places. Personally, I still don't really like that, it can easily confuse users/admins. 

The current situation, with just a single host_injection directory, also has some drawbacks, e.g. in case you want to do site installations on a local CVMFS repo, while having MPI/GPU libraries on a local disk. You can do it by doing some symlink trickery, but that's not ideal either.

Anyway, since the default behavior won't change, I was okay with adding the possibility for overriding the software installation prefix.

init/modules/EESSI/2023.06.lua
bedroge: Do we currently even support caches for local installation? Caspar and I looked into that a bit yesterday, but we couldn't find anything. We plan to cover this in the tutorial, but we should indeed also add documentation for this. With Caspar's proposed change, sites can easily call the `create_lmodrc.py` from our repo to generate the `lmodrc.lua` file for their local stack (the EESSI module will then add it to $LMOD_RC), and then it should just be a matter of calling for instance  https://github.com/EESSI/filesystem-layer/blob/main/scripts/update_lmod_caches.sh to actually generate the cache files.

init/modules/EESSI/2023.06.lua
casparvl: Happy to hear you've come around - I do feel that at least having the _option_ as a site to configure it is something we should offer. Whether a site uses it, or prefers redirecting the host-injections symlink (or making e.g. `/cvmfs/software.eessi.io/host_injections/2025.06` a symlink, as @ocaisa suggested on chat) are then just a choice that every site has.

I'd personally like to use the environment variable. And I'd _probably_ set it in an internal lmod hook, so that it's set at the moment users load the EESSI module - that way users can't unset it either.

Companion PR for EESSI-extend is here https://github.com/EESSI/software-layer-scripts/pull/235

I'll make issues for the rest. I think we should do the documentation updates once we've actually taken the whole new functionality for practical test - I'm hoping the idea is more clear after the webinar :)

casparvl: > We should also update the dev.eessi.io workflow to use this.

And yes, if this works well, I think it's nice to make dev.eessi.io use this, as it is essentially also 'building on top' of EESSI.

casparvl: Docs issue: https://github.com/EESSI/docs/issues/768
software-layer-scripts issue to update dev.eessi.io workflow https://github.com/EESSI/software-layer-scripts/issues/236

casparvl: Minimal docs update here https://github.com/EESSI/docs/pull/769 (we should still have more extensive docs on how to build on top of EESSI as a site, but that's what https://github.com/EESSI/docs/issues/768 is for).

init/modules/EESSI/2023.06.lua
```diff
@@ -134,7 +134,16 @@ local eessi_modules_subdir = pathJoin("modules", "all")
...
 -- /cvmfs/software.eessi.io/versions/<EESSI_VERSION>/software/linux/x86_64/amd/zen3/modules/all
 local eessi_module_path = pathJoin(eessi_software_path, eessi_modules_subdir)
-local eessi_site_software_path = string.gsub(eessi_software_path, "versions", "host_injections")
+local eessi_site_software_path
+-- If EESSI_SITE_SOFTWARE_PREFIX is defined, replace /cvmfs/software.eessi.io (or more generally EESSI_CVMFS_REPO)
+-- by that prefix. This ensures that the directory still contains the os/vendor/arch/micro-arch/accelerator etc
+-- If it is not defined, default to a site installation prefix under host_injections
+site_prefix = os.getenv("EESSI_SITE_SOFTWARE_PREFIX")
+if site_prefix then
+    eessi_site_software_path = string.gsub(eessi_software_path, os.getenv("EESSI_CVMFS_REPO"), site_prefix)
```
bedroge:
```suggestion
    eessi_site_software_path = string.gsub(eessi_software_path, eessi_repo, site_prefix)
```

init/modules/EESSI/2023.06.lua
```diff
@@ -210,8 +225,17 @@ end
 -- prepend the site module path last so it has priority
 prepend_path("MODULEPATH", eessi_site_module_path)
 eessiDebug("Adding " .. eessi_site_module_path .. " to MODULEPATH")
-if isDir(eessi_module_path_accel) then
+
+-- If EESSI_SITE_SOFTWARE_PREFIX is defined, replace /cvmfs/software.eessi.io (or more generally EESSI_CVMFS_REPO)
+-- by that prefix to get the site accelerator path. This ensures that the directory still contains the 
+-- os/vendor/arch/micro-arch/accelerator etc. If it is not defined, default to a site installation prefix under
+-- host_injections
+if site_prefix then
```
casparvl: I forgot this earlier, but we should implement the same logic for the site accelerator path.

init/modules/EESSI/2023.06.lua
```diff
@@ -210,8 +225,17 @@ end
...
+    eessi_module_path_site_accel = string.gsub(eessi_module_path_accel, eessi_repo, site_prefix)
+else
     eessi_module_path_site_accel = string.gsub(eessi_module_path_accel, "versions", "host_injections")
+end
+if isDir(eessi_module_path_site_accel) then
```
casparvl: I've changed the condition here: the original would effectively test "Is there an accelerator path in the _eessi_ prefix" and then add the accelerator path in the site prefix. That's totally weird, there's no reason why the addition of an accelerator path for a site should depend on the existence of that same accelerator path in EESSI. I don't see anything wrong with a site extending EESSI by e.g. building software for 2025.06 for a CUDA CC 11.0 target (which we don't support upstream).

Note that I'm not 100% sure that this (e.g. adding a CC11.0 target build) is totally supported with _only_ this change (e.g. we still need to check in line 233 that `eessi_module_path_accel` is defined in order to construct the right site accelerator path - and that might not be defined if the dir doesn't exist, I'm not sure), but it's a step in the right direction.

init/eessi_environment_variables
```diff
@@ -132,26 +148,28 @@ if [ -d $EESSI_PREFIX ]; then
...
           fi
 
-          if [ -d ${EESSI_ACCEL_SOFTWARE_PATH}/${EESSI_ACCELERATOR_TARGET}/${EESSI_MODULE_SUBDIR} ]; then
-              export EESSI_MODULEPATH_ACCEL=${EESSI_ACCEL_SOFTWARE_PATH}/${EESSI_ACCELERATOR_TARGET}/${EESSI_MODULE_SUBDIR}
+          if [ -d "${EESSI_ACCEL_SOFTWARE_PATH}/${EESSI_ACCELERATOR_TARGET}/${EESSI_MODULE_SUBDIR}" ]; then
+              export EESSI_MODULEPATH_ACCEL="${EESSI_ACCEL_SOFTWARE_PATH}/${EESSI_ACCELERATOR_TARGET}/${EESSI_MODULE_SUBDIR}"
               show_msg "Using ${EESSI_MODULEPATH_ACCEL} as additional directory (for accelerators) to be added to MODULEPATH."
-              export EESSI_SITE_MODULEPATH_ACCEL=${EESSI_SITE_ACCEL_SOFTWARE_PATH}/${EESSI_ACCELERATOR_TARGET}/${EESSI_MODULE_SUBDIR}
+          fi
+          if [ -d "${EESSI_SITE_ACCEL_SOFTWARE_PATH}/${EESSI_ACCELERATOR_TARGET}/${EESSI_MODULE_SUBDIR}" ]; then
```
casparvl, line 171: Change this in sync with the change in `2023.06.lua`, which now conditionally adds the site accerator path if _that_ accelerator path exists - ie this check:

```
if isDir(eessi_module_path_site_accel) then
```

init/modules/EESSI/2023.06.lua
```diff
@@ -210,11 +225,23 @@ end
...
+-- Note that we need the eessi_module_path_accel to construct either of these site installation accelerator paths
+if eessi_module_path_accel then
+    if site_prefix then
+        eessi_module_path_site_accel = string.gsub(eessi_module_path_accel, eessi_repo, site_prefix)
+    else
+        eessi_module_path_site_accel = string.gsub(eessi_module_path_accel, "versions", "host_injections")
+    end
+    if isDir(eessi_module_path_site_accel) then
+        setenv("EESSI_SITE_MODULEPATH_ACCEL", eessi_module_path_site_accel)
+        prepend_path("MODULEPATH", eessi_module_path_site_accel)
```
bedroge, line 242: We should probably also add an accelerator lmod dir to $LMOD_RC, such that sites can generate an Lmod cache for these installations?

bedroge: We add those accelerator caches as part of the CPU caches, see https://github.com/EESSI/filesystem-layer/blob/main/scripts/update_lmod_caches.sh#L48. So my question did not make sense.

bedroge (approved):
Lgtm

PR #240, opened 2026-05-29
Small change in EESSI-extend comment to allow redeployment on RISC-V

bedroge:
The settings in the EESSI-extend module of the RISC-V repo are outdated, because it was never rebuilt after the initial installation. This leads to issues with for instance the CPP search path. Fixing it by forcing a redeploy here. In the future we should just always rebuild it for the RISC-V repo too.


ocaisa (approved):
LGTM

ocaisa (approved):
LGTM

PR #242, opened 2026-06-02
add pre-configure hook to adjust configure options for Zoltan when ParMETIS is not a dependency

boegel:
I tested this manually, and confirmed it fixes the problem that occurred in https://github.com/EESSI/software-layer/pull/1515, where building Zoltan failed with:
```
fatal error: parmetis.h: No such file or directory
```

Also, configure output of Zoltan now shows this, as expected:
```
checking whether to use parmetis... no
```


ocaisa (dismissed):
LGTM

ocaisa (approved):
Also tested it, works as intended!

PR #243, opened 2026-06-02
Archdetect riscv update

hugomeiland:
This update will treat the isa line from a riscv cpu (e.g. rv64imafdcv_zicbom_zicboz_zicntr_zicond_zicsr_zifencei_zihintpause_zihpm_zfh_zfhmin_zca_zcd_zba_zbb_zbc_zbs_zkt_zve32f_zve32x_zve64d_zve64f_zve64x_zvfh_zvfhmin_zvkt_sscofpmf_sstc_svinval_svnapot_svpbmt) and sed it into seperate flags which can be matched as relevant (!) features, the same way we do for x86_64. This should allow cpu's with the same relevant features, but from different vendors to be matched to the same cpu paths.


bot/build.sh
```diff
@@ -271,6 +271,22 @@ fi
...
+        if [[ "${optarch}" == "GENERIC" ]]; then
+            INSTALL_SCRIPT_ARGS+=("--generic")
+        else
+            export EASYBUILD_OPTARCH="${optarch}"
+            echo "bot/build.sh: EASYBUILD_OPTARCH='${EASYBUILD_OPTARCH}' (from RISC-V optarch map)"
+        fi
+    else
+        echo "bot/build.sh: ERROR: no RISC-V optarch mapping for '${EESSI_SOFTWARE_SUBDIR_OVERRIDE}'" >&2
+        exit 1
+    fi
```
bedroge: I think it would be better/nicer if this is done by EESSI-extend. That way, both the bot and EESSI users doing builds on top of EESSI will get the right optimization flags when loading EESSI-extend. We sort of do the same for NVIDIA and AMD compute capabilities (https://github.com/EESSI/software-layer-scripts/blob/main/EESSI-extend-easybuild.eb#L216).

hugomeiland: Good call — done in d6d5ad1.

RISC-V `EASYBUILD_OPTARCH` is now set from `init/arch_specs/eessi_riscv_optarch.map` when loading **EESSI-extend** (same idea as CUDA/AMD compute capabilities): if `EESSI_SOFTWARE_SUBDIR` is `riscv64/*` and `EASYBUILD_OPTARCH` is not already set (e.g. `--generic`), the module looks up the map and `setenv`s it.

Removed the duplicate wiring from `bot/build.sh` and `EESSI-install-software.sh`. The shell helper `init/eessi_riscv_optarch.sh` stays for tests / optional direct use.

So both the bot and users extending EESSI get the mapped `-march`/`-mtune`/`-mcpu` flags via `module load EESSI-extend`.

init/arch_specs/eessi_arch_riscv.spec
```diff
@@ -1,7 +1,12 @@
...
 # CPU vendors: SiFive (0x489), Spacemit (0x710)
+# Spec lines must not use parentheses in trailing comments: update_arch_specs evals each line.
 
 # Software path in EESSI 	| Vendor ID 	| List of defining CPU features
-"riscv64/sifive/p550"		"0x489"		"rv64imafdch_zicsr_zifencei_zba_zbb_sscofpmf"	# HiFive Premier P550
-"riscv64/spacemit/x60"		"0x710"		"rv64imafdcv_sscofpmf_sstc_svpbmt_zicbom_zicboz_zicbop_zihintpause"	# Banana Pi F3
-"riscv64/spacemit/x60-k6.6"	"0x710"		"rv64imafdcv_zicbom_zicboz_zicntr_zicond_zicsr_zifencei_zihintpause_zihpm_zfh_zfhmin_zca_zcd_zba_zbb_zbc_zbs_zkt_zve32f_zve32x_zve64d_zve64f_zve64x_zvfh_zvfhmin_zvkt_sscofpmf_sstc_svinval_svnapot_svpbmt"	# Banana Pi F3 k6.6
+"riscv64/generic/rva20u64"	""		"rv64imafdc"
+"riscv64/generic/rva22u64"	""		"rv64imafdc zfhmin"
+"riscv64/generic/rva23u64"	""		"rv64imafdciv zfhmin"
```
HaoZeke: `rv64imafdciv` has `i` twice, and the kernel prints the single-letter block in canonical order ([1], "ISA string ordering rules"; the canonical order itself is Appendix C of the ISA manual [2]), so `/proc/cpuinfo` can never produce `...dciv`.
Should be `rv64imafdcv` I think.
Doesn't matter anyway right now because of the vendor gate, but it'll be a silent never-match once that's fixed.

[1]: https://www.kernel.org/doc/html/latest/arch/riscv/uabi.html
[2]: https://riscv.github.io/riscv-isa-manual/snapshot/spec/#app:naming

init/arch_specs/eessi_riscv_optarch.map
```diff
@@ -0,0 +1,35 @@
...
+#   -march (e.g. Ubuntu 24.04 gcc-riscv64 13.x / clang 18).
+# - Omit -mtune=generic: unknown to some RISC-V GCC builds.
+# - Omit S-mode / privileged extensions from -march (sscofpmf, sstc, sv*, ...):
+#   they are not needed for userspace codegen and break older Clang.
+# - No SpacemiT -mtune/-mcpu name upstream yet.
+# - Spec lines must not use parentheses in trailing comments.
+
+riscv64/generic			GENERIC
+riscv64/generic/rva20u64	GCC:-march=rv64gc -mabi=lp64d;Clang:-march=rv64gc -mabi=lp64d;LLVM:-march=rv64gc -mabi=lp64d
+riscv64/generic/rva22u64	GCC:-march=rv64gc_zba_zbb_zbs_zfhmin_zicbom_zicbop_zicboz -mabi=lp64d;Clang:-march=rv64gc_zba_zbb_zbs_zfhmin_zicbom_zicbop_zicboz -mabi=lp64d;LLVM:-march=rv64gc_zba_zbb_zbs_zfhmin_zicbom_zicbop_zicboz -mabi=lp64d
```
HaoZeke: The `-march` here looks right to me (RVA22U64 implies zba/zbb/zbs/zicbo* [1]), but the spec entry for the same subdir only asks for `rv64imafdc` + `zfhmin`, so wouldn't a host with zfhmin and no zbs get labeled rva22u64 and receive binaries it can't run?
Might be worth a test that every extension in a subdir's `-march` is required by its spec line; the vendored targets already pass that, it's only the profile entries.

[1]: https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva22u64_profile

init/eessi_archdetect.sh
```diff
@@ -154,6 +154,12 @@ cpupath(){
     fi
 
     local cpu_flags=$(get_cpuinfo "$cpu_flag_tag")
+    if [ "${machine_type}" == "riscv64" ]; then
+	# RISC-V ISA strings use '_' as extension separators.
+	# Convert them to space-separated feature tokens so they
+	# can be matched like x86 CPU flags.
+        cpu_flags=${cpu_flags//_/ }    
+    fi
```
HaoZeke, line 187: While the format is open: the leading `rv64imafdch` blob still matches as one token, so a superset host (say `rv64imafdcvh`) fails a spec asking `rv64imafdch`.
Splitting the blob into per-letter tokens would make the base extensions individually matchable, which I think is what the description is going for?
If the expansion happens on both sides at match time, the spec files keep the readable concatenated form, something like

```bash
riscv_expand_base(){
    # rv64imafdch -> rv64 i m a f d c h ; every other token passes through
    local out="" tok
    for tok in "$@"; do
        if [ "${tok#rv64}" != "${tok}" ]; then
            out="${out} rv64 $(printf '%s' "${tok#rv64}" | sed 's/./& /g')"
        else
            out="${out} ${tok}"
        fi
    done
    printf '%s' "${out}"
}

# here, for the host flags:
cpu_flags=$(riscv_expand_base ${cpu_flags})

# and at the match, for the spec side:
check_allinfirst "${cpu_flags[*]}" $(riscv_expand_base ${arch_spec[2]})
```

so a superset host matches any subset spec without the spec files caring how the kernel concatenated the letters.
Also trailing whitespace on this line, and the comment block is tab-indented against the file.

EESSI-extend-easybuild.eb
```diff
@@ -223,6 +223,42 @@ if (easybuild_amdgcn_compute_capabilities ~= nil) then
...
+          val = string.gsub(val, "^%s+", "")
+          val = string.gsub(val, "%s+$", "")
+          optarch = val
+          break
+        end
+      end
+    end
+    map_fh:close()
+    if optarch == nil or optarch == "" then
+      LmodError("No RISC-V EASYBUILD_OPTARCH mapping for '" .. eessi_software_subdir .. "' (add it to " .. map_file .. ")")
```
HaoZeke: This makes a missing map line fail `module load` for everyone on that target, including site installs with their own optarch policy.
Maybe warn and fall through to the default like the CUDA capabilities do?

.github/workflows/tests_riscv_optarch.yml
```diff
@@ -0,0 +1,85 @@
...
+          # Host is x86_64: use x86_64 compat, but modules from the RISC-V repo.
+          export EESSI_CPU_FAMILY_OVERRIDE=x86_64
+          export EESSI_SOFTWARE_SUBDIR_OVERRIDE=riscv64/generic
+          export EESSI_VERSION_OVERRIDE=2025.06-001
+
+          module load "EESSI/${EESSI_VERSION}"
+          echo "MODULEPATH=${MODULEPATH}"
+
+          # foss/2025b → GCC/14.3.0; lfoss/2025b → llvm-compilers/20.1.8
+          module load GCC/14.3.0
```
HaoZeke: Two things maybe worth a comment in the file: the module versions are pinned, so the next dev.eessi.io rebuild breaks this, and `-fsyntax-only` never reaches the assembler, so the job name overclaims a bit.
The triple override combo also skips the init scripts' own consistency checks; one line saying it's deliberate would help the next reader.

init/arch_specs/eessi_arch_riscv.spec
```diff
@@ -1,7 +1,12 @@
 # RISC-V CPU architecture specifications (see https://github.com/riscv/learn?tab=readme-ov-file#open-risc-v-implementations)
 # CPU vendors: SiFive (0x489), Spacemit (0x710)
+# Spec lines must not use parentheses in trailing comments: update_arch_specs evals each line.
 
 # Software path in EESSI 	| Vendor ID 	| List of defining CPU features
-"riscv64/sifive/p550"		"0x489"		"rv64imafdch_zicsr_zifencei_zba_zbb_sscofpmf"	# HiFive Premier P550
-"riscv64/spacemit/x60"		"0x710"		"rv64imafdcv_sscofpmf_sstc_svpbmt_zicbom_zicboz_zicbop_zihintpause"	# Banana Pi F3
-"riscv64/spacemit/x60-k6.6"	"0x710"		"rv64imafdcv_zicbom_zicboz_zicntr_zicond_zicsr_zifencei_zihintpause_zihpm_zfh_zfhmin_zca_zcd_zba_zbb_zbc_zbs_zkt_zve32f_zve32x_zve64d_zve64f_zve64x_zvfh_zvfhmin_zvkt_sscofpmf_sstc_svinval_svnapot_svpbmt"	# Banana Pi F3 k6.6
+"riscv64/generic/rva20u64"	""		"rv64imafdc"
```
HaoZeke: I don't think these three can ever match?
`cpupath()` compares vendors first [1] and only `aarch64` has an unknown-vendor fallback [2], so everything real gets an `mvendorid` [3] and an empty vendor field never survives the gate.

I think the `riscv64/sifive/u74-mc:riscv64/generic` in the u74-mc fixture [5] is also shown that way because of this.
Its isa line [4] tokenizes to

```
rv64imafdc zicntr zicsr zifencei zihpm zca zcd zba zbb
```

which contains everything the rva20u64 entry asks for (`rv64imafdc`, [6]), so feature-wise it should have matched.
Separately, the entry is a lot weaker than the actual RVA20U64 profile, which also mandates Zicsr, Zicntr, and Zifencei among others [7]; probably fine as a matching heuristic, but worth saying which one it's meant to be.

Maybe add a riscv fallback like the ARM `0x41` one [2], or drop the three entries until that exists?
There's also no `generic/rva*` fixture right now, so CI can't catch this.

[1]: https://github.com/EESSI/software-layer-scripts/blob/d6d5ad10f58e2b97746548e891b8f299ab78d281/init/eessi_archdetect.sh#L173
[2]: https://github.com/EESSI/software-layer-scripts/blob/d6d5ad10f58e2b97746548e891b8f299ab78d281/init/eessi_archdetect.sh#L129-L136
[3]: https://github.com/EESSI/software-layer-scripts/blob/d6d5ad10f58e2b97746548e891b8f299ab78d281/init/eessi_archdetect.sh#L116-L119
[4]: https://github.com/EESSI/software-layer-scripts/blob/d6d5ad10f58e2b97746548e891b8f299ab78d281/tests/archdetect/riscv64/sifive/u74-mc/starvision-Ubuntu24.cpuinfo
[5]: https://github.com/EESSI/software-layer-scripts/blob/d6d5ad10f58e2b97746548e891b8f299ab78d281/tests/archdetect/riscv64/sifive/u74-mc/starvision-Ubuntu24.all.output
[6]: https://github.com/EESSI/software-layer-scripts/blob/d6d5ad10f58e2b97746548e891b8f299ab78d281/init/arch_specs/eessi_arch_riscv.spec#L6
[7]: https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva20u64_profile

HaoZeke (changes requested):
The tokenization direction looks right to me and the vendored targets (p550, u74-mc, x60, x60-k6.6) are consistent between spec and map.

The generic profile entries are the problem: I don't think they can match at all under the vendor gate, one has a malformed token, and the specs are weaker than what the map compiles with; details inline.

On scope I'd split the optarch map + EESSI-extend Lua + its CI out, so the archdetect fix can merge while the optarch design gets its own discussion.
The map also has two parsers now, bash and Lua, which will drift.

init/arch_specs/eessi_arch_riscv.spec
hugomeiland: Agreed — fixed. `riscv64/generic/rva23u64` now uses `rv64imafdcv` (canonical single-letter order after letter expansion in `riscv_expand_base`).

Tip: 3d7f7e5 (`init/arch_specs/eessi_arch_riscv.spec`).

init/arch_specs/eessi_arch_riscv.spec
hugomeiland: Thanks — all three points addressed on the archdetect side (tip 3d7f7e5):

1. **Empty vendor = any.** `cpupath()` now treats an empty Vendor ID in the spec as matching any host `mvendorid`, so profile paths like `riscv64/generic/rva*` are not stuck behind the vendor gate.

2. **Floors = detectable official RVA*U64 mandates.** Matching tokens are the userspace subset that can appear in `/proc/cpuinfo isa`, including `zifencei` (with `zicsr`/`zicntr`). PMA/behaviour/EE contracts that never show up as cpuinfo tokens stay documented but omitted from floors.
   - RVA20U64: https://docs.riscv.org/reference/rva20-rvi20-rva22/v1.0/rva20.html
   - RVA22U64: https://docs.riscv.org/reference/rva20-rvi20-rva22/v1.0/rva22.html
   - RVA23U64: https://docs.riscv.org/reference/rva23/v1.0/rva23-profiles.html
   Current floors:
   - `rva20u64`: `rv64imafdc zicsr zicntr zifencei`
   - `rva22u64`: + `zihpm zihintpause zba zbb zbs zicbom zicbop zicboz zfhmin zkt`
   - `rva23u64`: + `v` / `zihintntl` / `zicond` / `zimop` / `zcmop` / `zcb` / `zfa` / `zawrs` / `zvfhmin` / `zvbb` / `zvkt` (and keeps RVA22 `zicbo*`)

3. **Fixtures.** Synthetic `generic/rva20|22|23u64` hosts are in the archdetect matrix. u74-mc `.all` is now `riscv64/sifive/u74-mc:riscv64/generic/rva20u64:riscv64/generic` as expected once empty-vendor + expansion land.

Vendor lines remain **full measured fixture ISA** (including S-mode tokens when advertised), not truncated profile floors — intentional for vendor-specific software trees.

init/eessi_archdetect.sh
hugomeiland: Agreed — done on this PR (tip 3d7f7e5; landed in 2e67c0d).

`riscv_expand_base()` expands compact `rv64*` blobs into per-letter tokens at match time on both host flags and the spec side (`rv64imafdch` → `rv64 i m a f d c h`; `g` → `imafd` before splitting). Multi-letter extensions pass through. Superset hosts now match subset specs without changing the readable concatenated form in `eessi_arch_riscv.spec`.

Also cleaned the trailing whitespace / indentation on that block.

init/arch_specs/eessi_riscv_optarch.map
hugomeiland: Good catch — two pieces:

1. **Archdetect floors (this PR / tip 3d7f7e5):** `riscv64/generic/rva22u64` (and rva20/rva23) now require the detectable official RVA*U64 userspace mandates that can appear in `/proc/cpuinfo` — including `zba`/`zbb`/`zbs`/`zicbo*`/`zfhmin` for RVA22 — so a host with only `zfhmin` and no `zbs` no longer labels as rva22u64. (Covered in more detail on the empty-vendor/floors thread.)

2. **map ⊆ spec (follow-up PR):** optarch map + the “every `-march` userspace token must be in the matching archdetect spec” test are intentionally split out of this PR. Already implemented locally on `archdetect-riscv-optarch` (`a4ce395`): `tests/riscv_optarch/test_eessi_riscv_optarch.sh` runs `check_map_subseteq_spec` (skips `GENERIC` and `-mcpu`-only rows like u74-mc). Will land with the optarch / EESSI-extend follow-up.

EESSI-extend-easybuild.eb
hugomeiland: Agreed — warn and fall through, same idea as CUDA capabilities.

Removed from this archdetect-only PR. On the local optarch follow-up (`archdetect-riscv-optarch`, `76c431b` / tip `a4ce395`): missing map entry is `LmodWarning` and leaves `EASYBUILD_OPTARCH` unset (site override / EasyBuild default), rather than `LmodError` failing `module load`. Covered by a smoke assert in `tests/riscv_optarch/`.

Also: production map parsing is Lua-only in EESSI-extend; bash helpers are test-oracle only so the two parsers don’t drift in production.

.github/workflows/tests_riscv_optarch.yml
hugomeiland: Agreed on all three — comments are in the follow-up branch (workflow not in this PR anymore).

On `archdetect-riscv-optarch` (`a4ce395`, `.github/workflows/tests_riscv_optarch.yml`):
- Step/job text notes modules are **pinned** (GCC/14.3.0, llvm-compilers/20.1.8) and will need bumps when the RISC-V stack rebuilds.
- Documents that we use **`-fsyntax-only`** on purpose: under qemu on x86 the host `as` rejects `-march=rv64*`, so we never claim a full assemble/`-c` compile.
- Notes that the `EESSI_*_OVERRIDE` combo is **deliberate** for CI (x86 compat host + RISC-V modules) and skips the init scripts’ usual consistency path by design.

init/arch_specs/eessi_riscv_optarch.map
julianmorillo: One comment here: why the distinction Clang vs LLVM?

hugomeiland: This is due to EasyBuild, not to Clang/LLVM as projects. EasyBuild keeps two separate toolchain families (`COMPILER_FAMILY`): legacy `Clang` vs `LLVM` (`llvm-compilers` / `lfoss`). `--optarch` matches on that family name, so the map needs both `Clang:` and `LLVM:` keys (even when the flags are the same on RISC-V today). EESSI-extend just sets the full EasyBuild multi-compiler string; it does not invent the split.

.github/workflows/tests_archdetect.yml
```diff
@@ -19,17 +19,19 @@ jobs:
...
           - aarch64/nvidia/grace/Jureca-Rocky95
           - aarch64/google/axion/GCP-axion
-          - riscv64/sifive/premier-Ubuntu24
+          - riscv64/sifive/p550/premier-Ubuntu24
+          - riscv64/sifive/u74-mc/starvision-Ubuntu24
           - riscv64/spacemit/bananaf3-Armbian
           - riscv64/spacemit/bananaf3-k6.6
+          - riscv64/generic/rva20u64/synthetic
+          - riscv64/generic/rva22u64/synthetic
+          - riscv64/generic/rva23u64/synthetic
```
julianmorillo:
```suggestion
          - riscv64/generic
          - riscv64/rva20u64
          - riscv64/rva22u64
          - riscv64/rva23u64
```
- I would keep paths as shorter as possible (while meaningful). 
- Keywords "generic" and "synthetic" do not provide extra information in this context. 
- I would use this flattened categorization where `riscv64/generic` is left for the most generic case (rv64gc) and each profile has its own first level cathegory.
As this would imply modifications in other files, I'll let this to be clarified first before going on with more comments.

hugomeiland: Agreed — flattened in tip `62763c9`.

Profile paths are now:
- `riscv64/rva20u64`
- `riscv64/rva22u64`
- `riscv64/rva23u64`

`riscv64/generic` stays as the always-present catch-all from `eessi_archdetect.sh` (not a nested profile). Fixture files moved to `tests/archdetect/riscv64/rva*.{cpuinfo,output,all.output}` (dropped the `generic/…/synthetic` layout). CI matrix and vendor `.all` expectations updated accordingly.

Happy for you to continue the review from here.

init/arch_specs/eessi_riscv_optarch.map
julianmorillo: Any reason to not use here the profile name for the `-march` option? I.e. use `-march=rva22u64` instead of  `-march=rv64gc_zba_zbb_zbs_zfhmin_zicbom_zicbop_zicboz`. It would benefit readability and it would be less error prone, I would say.

hugomeiland: Good suggestion, thanks. We tested this on the actual RISC-V target (RV2) with EESSI `GCCcore/14.3.0`, and profile aliases are not accepted yet:

- `-march=rva20u64` -> fail
- `-march=rva22u64` -> fail
- `-march=rva23u64` -> fail
- explicit `rv64...` ISA strings work

So for now the map keeps explicit `rv64*` `-march` values for compatibility across the current toolchains. Once profile-name support is reliably available in our GCC/Clang baseline, we can switch to profile names for readability.

boegel: I would definitely include that info in a comment in this file itself, to explain why we're not using the shorthands just yet?

hugomeiland: that is in the first line of the note above the code.....

julianmorillo: > Good suggestion, thanks. We tested this on the actual RISC-V target (RV2) with EESSI `GCCcore/14.3.0`, and profile aliases are not accepted yet:
> 
>     * `-march=rva20u64` -> fail
> 
>     * `-march=rva22u64` -> fail
> 
>     * `-march=rva23u64` -> fail
> 
>     * explicit `rv64...` ISA strings work
> 
> 
> So for now the map keeps explicit `rv64*` `-march` values for compatibility across the current toolchains. Once profile-name support is reliably available in our GCC/Clang baseline, we can switch to profile names for readability.

Oh, I see... I thought this was already tested. Then let's keep the explicit values.

hugomeiland: also this code is no longer in this PR; we agreed to split the PR earlier....

julianmorillo (approved):

PR #248, opened 2026-06-05
Add a fallback version for EasyBuild in the EESSI-extend module

ocaisa:
Without this change, you cannot run
```
module show EESSI-extend
```
unless you have an EasyBuild module loaded.


bedroge (approved):
Lgtm

PR #250, opened 2026-06-09
Use correct environment variable in hook that checks accelerator installation paths

bedroge:
The message said that you should do `EESSI_ACCELERATOR_TARGET=1 module load EESSI-extend`, but that does not work:
```
EESSI_ACCELERATOR_TARGET=1 module load EESSI-extend
Lmod Warning:  Incorrect value for $EESSI_ACCELERATOR_TARGET: 1 
```

The line above has the correct env v ar, it should be `EESSI_ACCELERATOR_INSTALL`.

Also fixed the unset, as `unset VAR=value` does not work.


ocaisa (approved):
Whoops, thanks for catching this

PR #252, opened 2026-06-10
Make cfg_files.sh more robut in bash environment

laraPPr:
This causes issues in environment where `shopt -s nullglob`. Which is the case for our local bot where we use cfg_files.sh. This small change will make the funtion a bit more robust.


boegel (approved):
lgtm

PR #254, opened 2026-06-11
Skip tests that require gpus when crosscompiling Siesta

laraPPr:
Edit by Caspar: A subset of Siesta tests require GPUs to run. These tests fail when cross-compiling. This PR skips the tests if no GPU is present. The issue was originally hit here: https://github.com/EESSI/software-layer/pull/1506#issuecomment-4660833530

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -1707,19 +1707,22 @@ def pre_test_hook_ignore_failing_tests_OpenBabel_a64fx(self, *args, **kwargs):
 
 def pre_test_hook_Siesta_ignore_failure_with_crosscompilation(self, *args, **kwargs):
     """
-    Ignore failing tests when crosscompiling without gpu pressent.
+    Ignore failing tests when crosscompiling without gpu present.
     """
     if self.name == 'Siesta': 
         if self.version in ['5.4.2']:
-            cuda_cc = build_option('cuda_compute_capabilities')
-            if cuda_cc and not get_gpu_info(): 
-                failing_tests=[
-                    "Solvers-si-qdot-elsi-elpa-gpu_mpi4_omp1", # runs cuda get device
-                    "Solvers-si-qdot-elsi-elpa-1stage-gpu_mpi4_omp1", # runs cuda get device
-                    "Solvers-si-qdot-elpa-native-gpu_mpi4_omp1", # runs cuda get device
-                    "Solvers-si-qdot-elpa-native-1stage-gpu_mpi4_omp1", # runs cuda get device
-                ]
-                self.cfg['testopts'] = self.cfg['testopts'][:-1] + "|" + "|".join(failing_tests) + "'"
+            if 'CUDA' in self.cfg['versionsuffix']: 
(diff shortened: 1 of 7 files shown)


eb_hooks.py
```diff
@@ -1705,6 +1705,23 @@ def pre_test_hook_ignore_failing_tests_OpenBabel_a64fx(self, *args, **kwargs):
...
+        if self.version in ['5.4.2']:
+            cuda_cc = build_option('cuda_compute_capabilities')
+            if cuda_cc and not get_gpu_info(): 
+                failing_tests=[
+                    "Solvers-si-qdot-elsi-elpa-gpu_mpi4_omp1", # runs cuda get device
+                    "Solvers-si-qdot-elsi-elpa-1stage-gpu_mpi4_omp1", # runs cuda get device
+                    "Solvers-si-qdot-elpa-native-gpu_mpi4_omp1", # runs cuda get device
+                    "Solvers-si-qdot-elpa-native-1stage-gpu_mpi4_omp1", # runs cuda get device
+                ]
+                self.cfg['testopts'] = self.cfg['testopts'][:-1] + "|" + "|".join(failing_tests) + "'"
```
boegel: You're making a lot of assumptions here, maybe it's better to use a regular expression here to keep the part that's already there in the `-E '...'`, and only add to it:

```python
>>> testopts
"--output-on-failure -E 'foo|bar'"
>>>
>>>
>>> re.sub(r"-E '(.*)'", r"-E '\1|one|two'", testopts)
"--output-on-failure -E 'foo|bar|one|two'"
```

eb_hooks.py
```diff
@@ -1705,6 +1705,23 @@ def pre_test_hook_ignore_failing_tests_OpenBabel_a64fx(self, *args, **kwargs):
         self.cfg['testopts'] = "|| echo ignoring failing tests"
 
 
+def pre_test_hook_Siesta_ignore_failure_with_crosscompilation(self, *args, **kwargs):
+    """
+    Ignore failing tests when crosscompiling without gpu pressent.
+    """
+    if self.name == 'Siesta': 
+        if self.version in ['5.4.2']:
```
boegel, line 1739: Also check for `CUDA` in `self.cfg['versionsuffix']`?

We should only skip these tests for CUDA easyconfigs of Siesta

software.eessi.io
```diff
@@ -0,0 +1 @@
+/cvmfs/software.eessi.io
```
casparvl: Is this file/link supposed to be here...?

laraPPr: I don't know where it came from. I'll fix it but I also need to make updates with Kenneths suggestions so will do it then.

casparvl (changes requested):
Since you've made changes to the `eb_hooks.py` _after_ testing: can you re-add the easystack file, test the build again, and remove the easystack file again? (to prove that this version of `eb_hooks.py` still works)

casparvl (approved):

PR #259, opened 2026-06-24
Add rompi/2025a as a top-level toolchain for EasyBuild >= 5.3.1

zerefwayne:


eb_hooks.py
casparvl: You need to adapt this, or it will never make it to the `elif` case (since this is _also_ true for 5.3.1)

```suggestion
if EASYBUILD_VERSION == '5.3.0':
```

casparvl (approved):
As discussed, there was no point in having this for 5.3.0 already, since 5.3.0 contained _zero_ easyconfigs based on `rocm-compilers`. We probably had it to be able to do `--from-commit` stuff, but since that never made it into an actual easystack in production, we don't need that anymore. So, we just bump directly to `rompi` since that is available in 5.3.1.

PR #261, opened 2026-07-08
Fix and move copy_build_log function

laraPPr:
We would also like to use this funtion in our local build bot at UGent so moving it helps us.

There is also an issue with the current funtion. It is not printing the job_id:
```
Context from which build log was copied:

- original path of build log: /tmp/eb-3wcdvarr/easybuild-4epar1do.log

- working directory: /eessi_bot_job

- Slurm job ID:

- EasyBuild version: 5.3.1

- easystack file: easystacks/software.eessi.io/2025.06/eessi-2025.06-eb-5.3.1-2025b.yml

```

### DIFF
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -17,37 +17,6 @@ display_help() {
   echo "  --skip-cuda-install    -  disable installing a full CUDA SDK in the host_injections prefix (e.g. in CI)"
 }
 
-function copy_build_log() {
-    # copy specified build log to specified directory, with some context added
-    build_log=${1}
-    build_logs_dir=${2}
-
-    # also copy to build logs directory, if specified
-    if [ ! -z "${build_logs_dir}" ]; then
-        log_filename="$(basename ${build_log})"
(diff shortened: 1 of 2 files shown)


ocaisa (approved):
LGTM

EESSI-install-software.sh
```diff
@@ -17,37 +17,6 @@ display_help() {
...
-        mkdir -p $(dirname ${build_log_path})
-        cp -a ${build_log} ${build_log_path}
-        chmod 0644 ${build_log_path}
-
-        # add context to end of copied log file
-        echo >> ${build_log_path}
-        echo "Context from which build log was copied:" >> ${build_log_path}
-        echo "- original path of build log: ${build_log}" >> ${build_log_path}
-        echo "- working directory: ${PWD}" >> ${build_log_path}
-        echo "- Slurm job ID: ${SLURM_OUT}" >> ${build_log_path}
```
ocaisa, line 43: Indeed `SLURM_OUT` does not exist, nice catch

PR #231, opened 2026-05-14
Allow for compiler+accelerator specific MPI overrides

ocaisa:
Alternative to #230 where we focus only on the potential need for CUDA/ROCm variants.

This also opens the door to other types of variants (but the options here would be multiplicative so I haven't included that until we hit a need for it).

PR #262, opened 2026-07-08
Don't fail hooks if `CUDA-Samples` version is not recognised

ocaisa:
Will wait for #231 before proceeding with this

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -1130,21 +1130,23 @@ def pre_configure_hook_BLIS(self, *args, **kwargs):
 
 def pre_configure_hook_CUDA_Samples_test_remove(self, *args, **kwargs):
     """skip immaTensorCoreGemm in CUDA-Samples for compute capability 7.0."""
-    if self.name == 'CUDA-Samples' and self.version in ['12.1']:
-        # Get compute capability from build option
-        cuda_caps = build_option('cuda_compute_capabilities')
-        # Check if compute capability 7.0 is in the list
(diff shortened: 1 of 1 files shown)


eb_hooks.py
```diff
@@ -1141,21 +1141,23 @@ def pre_configure_hook_BLIS(self, *args, **kwargs):
...
+                print_msg("Applying hook for CUDA-Samples %s with compute capability 7.0", self.version)
+                # local_filters is set by the easyblock, remove path to the Makefile instead
+                makefile_path = os.path.join(self.start_dir, 'Samples/3_CUDA_Features/immaTensorCoreGemm/Makefile')
+                if os.path.exists(makefile_path):
+                    remove_file(makefile_path)
+                    print_msg("Removed Makefile at %s to skip immaTensorCoreGemm build", makefile_path)
+                else:
+                    print_msg("Makefile not found at %s", makefile_path)
+        else:
+            print_warning(f"CUDA-Samples hook triggered but version {self.version} does not have a code path.")
```
ocaisa: Specifically, I was testing https://github.com/EESSI/software-layer/pull/1541 and this was failing due to the strictness. At least for that version of CUDA-Samples a hook is not required

ocaisa: I could just leave out the `else` altogether rather than generate doubt

ocaisa: Changed it to an info message

TopRichard (approved):
lgtm

laraPPr (approved):
lgtm

PR #268, opened 2026-07-21
Update all CI checks to also run for 2026.06

casparvl:
I'm expecting some of these to not work (yet). I'm guessing we need some very basic stuff to be in place first. There may even be some CUDA-related GPU things that will not work yet - but let's just see how far this gets.


boegel:
@bedroge still failing tests?

ocaisa:
One failure is our Lmod having problems with `set -e` (fixed upstream), the other is because there is no cc70 in 2026.06

bedroge:
The test checks if cc77 falls back to cc70, that doesn't work for 2026.06. That either requires:
- a change in the EESSI module: falling back to x0 is now hardcoded at https://github.com/EESSI/software-layer-scripts/blob/main/init/modules/EESSI/2023.06.lua#L213. For 2026.06 it would need a more explicit check and let any 75 < cc <80 fall back to 75.
- or shall I just override that check for 2026.06, since there is no 75 < cc <80 in practice anyway

The latter seems much easier to me.

bedroge:
@ocaisa is there any workaround for the `set -e` issue, or do we really have to update Lmod (that will take some time, would be nice to have the CI enabled for 2026.06)?

PR #269, opened 2026-07-23
In 2023.06 don't fail for missing PTX for CUDA install, allow overriding the easystacks directory, add support for Tegra

ocaisa:
When you only want specific CUDA versions, it can be very slow to install all the options


scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
bedroge: @ocaisa this probably needs to be updated?

ocaisa: Done

bedroge (approved):

PR #282, opened 2026-08-07
Restrict Perl hook to EESSI 2026.06, also apply it to Perl 5.38.0

bedroge:
Didn't want to use a too big hammer, as Perl 5.38.0 with SYSTEM toolchain is also part of EESSI 2025.06, and we didn't see any issues there. So I've restricted it to EESSI 2026.06 and specific Perl versions.


boegel (dismissed):
I don't think it would hurt to actually always do this, it's just setting a very specific environment variable, I don't see how that could cause trouble elsewhere (worst case, it leads to a longer `test` step for Perl before it fails?)

But only doing it in specific cases is fine by me too, so lgtm

I'm a bit puzzled why we only see this now, and not with earlier EESSI versions...

boegel (approved):
lgtm

PR #283, opened 2026-08-07
Fetch all source files before processing easystack file

bedroge:
This should make it fail early in case source tarballs cannot be downloaded. Note that if you're processing multiple easystack files, it will not fetch the sources for all of them. Doing that would require more changes to the code, and we're not doing that very often anyway: almost all PRs change only one easystack file, so this should already be a big improvement.

I'll give it a try with an actual easystack file.

### DIFF
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -430,13 +430,16 @@ else
             echo_green "All set, let's start installing some software with EasyBuild v${eb_version} in ${EASYBUILD_INSTALLPATH}..."
 
             if [ -f ${easystack_file} ]; then
-                echo_green "Downloading sources for easystack file ${easystack_file}..."
-                # run eb --help to determine if --fetch-all is supported (available since 5.3.0), otherwise fall back to --fetch
-                fetch_option="--fetch"
-                ${EB} --help | grep -q -e "fetch-all" && fetch_option="--fetch-all"
-                ${EB} ${fetch_option} --easystack ${easystack_file} --robot
-                if [ $? -ne 0 ]; then
-                    fatal_error "Could not download all required source files for this easystack file."
+                if [ -z $EESSI_SKIP_FETCH_EASYSTACK_SOURCES ]; then
+                    # run eb --help to determine if --fetch-all is supported (available since 5.3.0), otherwise fall back to --fetch
+                    fetch_option="--fetch"
+                    ${EB} --help | grep -q -e "fetch-all" && fetch_option="--fetch-all"
+                    echo_green "Downloading sources for easystack file ${easystack_file} using eb ${fetch_option}..."
+                    echo "(note: this step can be skipped by setting $EESSI_SKIP_FETCH_EASYSTACK_SOURCES to a non-empty value)"
+                    ${EB} ${fetch_option} --easystack ${easystack_file} --robot
+                    if [ $? -ne 0 ]; then
+                        fatal_error "Could not download all required source files for this easystack file."
+                    fi
                 fi
 
                 echo_green "Feeding easystack file ${easystack_file} to EasyBuild..."
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -430,6 +430,12 @@ else
             echo_green "All set, let's start installing some software with EasyBuild v${eb_version} in ${EASYBUILD_INSTALLPATH}..."
 
             if [ -f ${easystack_file} ]; then
+                echo_green "Downloading sources for easystack file ${easystack_file}..."
+                ${EB} --fetch-all --easystack ${easystack_file} --robot
+                if [ $? -ne 0 ]; then
+                    fatal_error "Could not download all required source files for this easystack file."
+                fi
+
                 echo_green "Feeding easystack file ${easystack_file} to EasyBuild..."
 
                 if [[ ${easystack_file} == *"/rebuilds/"* ]]; then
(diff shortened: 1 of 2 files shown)


EESSI-install-software.sh
```diff
@@ -430,6 +430,12 @@ else
             echo_green "All set, let's start installing some software with EasyBuild v${eb_version} in ${EASYBUILD_INSTALLPATH}..."
 
             if [ -f ${easystack_file} ]; then
+                echo_green "Downloading sources for easystack file ${easystack_file}..."
+                ${EB} --fetch-all --easystack ${easystack_file} --robot
```
boegel: `--fetch-all` is only supported from EasyBuild 5.3.0 onwards, do we need a version check here?

bedroge: Good point. I guess it's not very likely that we'll still be using older versions, but I've added a check by running `eb --help` and grepping for `--fetch-all` first. Otherwise it falls back to `--fetch`.

bedroge:
> Does this mean that if the sources for already existing software in the easystack file cannot be fetched, the build process will fail? so we need to fix the existing easyconfigs first?

I guess that could happen, theoretically, e.g. if we would clean up the sources for some already installed applications. But I don't think we've ever done that. Also, we usually stick to the latest EB release, so the easyconfigs should be quite up-to-date, and it should not really happen that we are trying to use outdated source URLs from an old easyconfig.

bedroge:
> Does this mean that if the sources for already existing software in the easystack file cannot be fetched, the build process will fail? so we need to fix the existing easyconfigs first?

As discussed in the support meeting, I've introduced an environment variable (`$EESSI_SKIP_FETCH_EASYSTACK_SOURCES`) that allows you to disable this step.

boegel:
> > Does this mean that if the sources for already existing software in the easystack file cannot be fetched, the build process will fail? so we need to fix the existing easyconfigs first?
> 
> I guess that could happen, theoretically, e.g. if we would clean up the sources for some already installed applications. But I don't think we've ever done that. Also, we usually stick to the latest EB release, so the easyconfigs should be quite up-to-date, and it should not really happen that we are trying to use outdated source URLs from an old easyconfig.

Download problems can also happen because upstream is down (like https://cmake.org right now, or https://ftpmirror.gnu.org/ having trouble for a while now).

But the "cached" downloads should largely prevent frequent trouble there, indeed...

EESSI-install-software.sh
```diff
@@ -436,6 +436,18 @@ else
             echo_green "All set, let's start installing some software with EasyBuild v${eb_version} in ${EASYBUILD_INSTALLPATH}..."
 
             if [ -f ${easystack_file} ]; then
+                if [ -z $EESSI_SKIP_FETCH_EASYSTACK_SOURCES ]; then
+                    # run eb --help to determine if --fetch-all is supported (available since 5.3.0), otherwise fall back to --fetch
+                    fetch_option="--fetch"
+                    ${EB} --help | grep -q -e "fetch-all" && fetch_option="--fetch-all"
+                    echo_green "Downloading sources for easystack file ${easystack_file} using eb ${fetch_option}..."
+                    echo "(note: this step can be skipped by setting $EESSI_SKIP_FETCH_EASYSTACK_SOURCES to a non-empty value)"
```
boegel: needs escape here, or single quotes, otherwise `$EESSI_SKIP_FETCH_EASYSTACK_SOURCES` will be expanded to a blank:
```suggestion
                    echo '(note: this step can be skipped by setting $EESSI_SKIP_FETCH_EASYSTACK_SOURCES to a non-empty value)'
```

EESSI-install-software.sh
```diff
@@ -436,6 +436,18 @@ else
...
+                    ${EB} ${fetch_option} --easystack ${easystack_file} --robot
```
boegel: In some situations, this is going to result in quite a bit of slowdown, even if all source files for the easyconfigs listed in the easystack file are already available in the cached downloads (`sourcepath`), since this will iterate over *all* required sources (due to `--robot`), including the ones for extensions...

I understand how prefetching of sources helps, but is the above really what we want?

Can we somehow extract the list of easyconfig files being *added* to the easystack file, and only call `--fetch-all` on those?

bedroge: That's a good idea. I wasn't sure if that was easy to implement, but with some help of a friend I got an awk command that could fetch the added lines for each easystack file from the PR diff file. Still feels a little bit brittle, but let me give it a try.

EESSI-install-software.sh
```diff
@@ -436,6 +436,28 @@ else
...
 
             if [ -f ${easystack_file} ]; then
+                if [ -z $EESSI_SKIP_FETCH_EASYSTACK_SOURCES ]; then
+                    # run eb --help to determine if --fetch-all is supported (available since 5.3.0), otherwise fall back to --fetch
+                    fetch_option="--fetch"
+                    ${EB} --help | grep -q -e "fetch-all" && fetch_option="--fetch-all"
+                    easystack_additions=$(mktemp --suffix=.yml)
+                    # use the PR diff file to find which items have been added to this particular easystack file
+                    # and dump that to a temporary easystack file that we can use to fetch the sources for only these added items
+                    awk -v file="${easystack_file}" '
```
boegel: Are we sure `awk` is available everywhere we need it to be?

bedroge: Both the compat layer and container have `awk`.

PR #284, opened 2026-08-07
Extend CUDA supported CC table, add CUDA and cuDNN host injections easystack for 2026.06

bedroge:
Extended the table using the instructions at https://github.com/EESSI/software-layer-scripts/blob/main/eb_hooks.py#L90.

Added a host injections easystack for 2026.06 CUDA+cuDNN.


casparvl (changes requested):
Checked the logs for the `2026.06` build. It fails to install the new CUDA and cuDNN you added here into `host-injections`. The reason is a mistake I made in https://github.com/EESSI/software-layer-scripts/pull/266 . It now uses

```
$TOPDIR/easystacks
```

As the directory to search for easystack files. That is ok for the deployed `install_cuda_and_libraries.sh`, but not for the one that is shipped with the GitHub repository, since there the easystacks are in `$TOPDIR/easystacks/$EESSI_VERSION`.

We can fix this by setting the `NVIDIA_EASYSTACKS_DIRECTORY` environment variable to `$TOPDIR/easystacks/$EESSI_VERSION`, right before `EESSI_install_software.sh` calls `install_cuda_and_libraries.sh`.

casparvl (approved):

PR #288, opened 2026-08-12
Fix CUDA sanity check for EasyBuild >=5.4.0

casparvl:
Since EB 5.4.0, the build dependencies are no longer available during the sanity check phase. This breaks our CUDA sanity check, which relies on cuobjdump being available.

This PR tries to promote CUDA to a regular dep temporarily, hoping that this will make the CUDA sanity check pass. A test on CUDA-Samples 12.9 shows that this is NOT the case. My AI friend thinks it's because the fake module that is generated is being generated based on data structures that are already available, so modifying the existing self.cfg instance has no effect.

Anyway, this is at least a description of the issue,and a starting point for the solution...

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -986,7 +986,12 @@ def pre_sanitycheck_hook_cuda(self, *args, **kwargs):
                 setattr(self, EESSI_SANITYCHECK_CUDA_ATTR, dep)
                 # Load CUDA module
                 self.modules_tool.load([dep['full_mod_name']])
-                print_msg(f"Loading CUDA module '{dep['full_mod_name']}', temporarily making available for the (CUDA) sanity check")
+                cuobjdump_path = shutil.which('cuobjdump')
+                cuobjdump_dir = os.path.dirname(cuobjdump_path)
+                self.cuobjdump_dir = cuobjdump_dir
+                os.environ['PATH'] = os.environ.get('PATH', '') + os.pathsep + cuobjdump_dir
+                print_msg(f"Adding location of cuobjdump_dir ({cuobjdump_dir}) to the PATH so that we can execute the CUDA sanity check")
+                self.modules_tool.unload([dep['full_mod_name']])
                 break
 
 
@@ -995,13 +1000,10 @@ def post_sanitycheck_hook_cuda(self, *args, **kwargs):
     Reverse the temporary CUDA dependency promotion from pre_sanitycheck_hook_cuda.
     """
     if hasattr(self, EESSI_SANITYCHECK_CUDA_ATTR):
-        cuda_dep = getattr(self, EESSI_SANITYCHECK_CUDA_ATTR)
-        # Remove from runtime dependencies
-        for dep in self.cfg['dependencies']:
-            if dep['name'] == 'CUDA':
-                self.modules_tool.unload([dep['full_mod_name']])
-                print_msg(f"Unloading CUDA module '{dep['full_mod_name']}', temporarily making available for the (CUDA) sanity check")
-                break
+        # Technically, we should probably remove the cuobjdump_dir from the path
+        # for now, we do nothing, let's first check this works.
+        # TODO: cleanup
+        # print_msg(f"Remove {self.cuobjdump_dir} from the PATH")
         delattr(self, EESSI_SANITYCHECK_CUDA_ATTR)
 
 
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -989,9 +989,9 @@ def pre_sanitycheck_hook_cuda(self, *args, **kwargs):
                 cuobjdump_path = shutil.which('cuobjdump')
                 cuobjdump_dir = os.path.dirname(cuobjdump_path)
                 self.cuobjdump_dir = cuobjdump_dir
+                self.modules_tool.unload([dep['full_mod_name']])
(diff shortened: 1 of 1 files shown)


eb_hooks.py
```diff
@@ -957,6 +958,55 @@ def post_module_hook_unsupported_module(self, *args, **kwargs):
...
+            if dep['name'] == 'CUDA':
+                # Store the dependency for removal in post hook
+                setattr(self, EESSI_SANITYCHECK_CUDA_ATTR, dep)
+                # Load CUDA module
+                self.modules_tool.load([dep['full_mod_name']])
+                cuobjdump_path = shutil.which('cuobjdump')
+                cuobjdump_dir = os.path.dirname(cuobjdump_path)
+                self.cuobjdump_dir = cuobjdump_dir
+                self.modules_tool.unload([dep['full_mod_name']])
+                os.environ['PATH'] = os.environ.get('PATH', '') + os.pathsep + cuobjdump_dir
```
ocaisa: Probably want to store the original value in the attribute and restore it when removing that attribute

casparvl: Yeah, that was a 'TODO' (see the post-hook). But... it doesn't work. I don't have time to look into it any further right now... If you're up for it, you can give it a go. Otherwise, I expect Bob will look into it somewhere next week - he hit it when trying to compile ESPResSo with CUDA support.

eb_hooks.py
```diff
@@ -957,6 +958,55 @@ def post_module_hook_unsupported_module(self, *args, **kwargs):
...
+        for dep in build_deps:
+            if dep['name'] == 'CUDA':
+                # Store the dependency for removal in post hook
+                setattr(self, EESSI_SANITYCHECK_CUDA_ATTR, dep)
+                # Load CUDA module
+                self.modules_tool.load([dep['full_mod_name']])
+                cuobjdump_path = shutil.which('cuobjdump')
+                cuobjdump_dir = os.path.dirname(cuobjdump_path)
+                self.cuobjdump_dir = cuobjdump_dir
+                self.modules_tool.unload([dep['full_mod_name']])
```
laraPPr, line 991: In the case of LAMMPS we also need a CUDA during th running of the tests. So this hook would not fix the failing sanity check fo LAMMPS.

laraPPr: Should there than be another hook for LAMMPS which loads CUDA in `pre_sanity_check_hook` and unloads in `post_sanity_check_hook`?

laraPPr: It seems like this is also not really loading the module. Because when I add this to the hooks without an unload the sanity check still fails because of missing CUDA.

ocaisa: This change should be doing this for any software that requires CUDA (so should already help with LAMMPS)

eb_hooks.py
```diff
@@ -957,6 +958,55 @@ def post_module_hook_unsupported_module(self, *args, **kwargs):
...
+    if cudaver and EASYBUILD_VERSION >= '5.4.0':
+        # Get the CUDA dependency info from builddependencies
+        build_deps = self.cfg.get_ref('builddependencies')
+        for dep in build_deps:
+            if dep['name'] == 'CUDA':
+                # Store the dependency for removal in post hook
+                setattr(self, EESSI_SANITYCHECK_CUDA_ATTR, dep)
+                # Load CUDA module
+                self.modules_tool.load([dep['full_mod_name']])
+                cuobjdump_path = shutil.which('cuobjdump')
```
laraPPr: shutil does not seem to be imported so I don't know how this even worked

eb_hooks.py
laraPPr: I have been trying this hook. but it does not seem like it works. All the tests during the sanity check keep failing. And I also still see missing cuobjdump error.

laraPPr: I got the tests working by removing this module unload. But the cuobjdump error is still their.

eb_hooks.py
ocaisa: Done in https://github.com/EESSI/software-layer-scripts/pull/288/commits/4a932f9f2695fe57f6736f6592994564a54c0eaf

eb_hooks.py
ocaisa: Fixed in https://github.com/EESSI/software-layer-scripts/pull/288/commits/4a932f9f2695fe57f6736f6592994564a54c0eaf

eb_hooks.py
ocaisa: Replaced in https://github.com/EESSI/software-layer-scripts/pull/288/commits/4a932f9f2695fe57f6736f6592994564a54c0eaf

ocaisa (approved):
I'm ok with this now, but I had a significant hand in the final version so I leave it to @laraPPr to review/deploy

boegel:
I'm surprised this worked in the first place with older EasyBuild versions, since build dependencies were never meant to be loaded during the sanity check (which should mimic the environment as a user would use the module).

It could be a side-effect of the changes made in:
- https://github.com/easybuilders/easybuild-framework/pull/5211

But that shouldn't have changed anything w.r.t. environment for the sanity check step...

Or perhaps it's related to this bug fix:
- https://github.com/easybuilders/easybuild-framework/pull/5185

laraPPr (approved):
lgtm

